PR #112, opened 2022-07-01
Feature/83/create who we are section

DeLettenhove:
Create who we are section of the guest home page.

image


src/containers/guest-home-page/WhoWeAre.js
```diff
@@ -0,0 +1,53 @@
...
+    maxWidth: '100%',
+  }
+}
+
+
+const WhoWeAre = () => {
+  const { t } = useTranslation()
+
+  return (
+    <Box sx={ style.container }>
```
LubomyrKashuba, line 35: use className 'section'. It is global class with marginBottom placed in index.css

DeLettenhove: Fixed.

tarasSanotskyi (approved):

PR #180, opened 2022-07-27
Add VideoBox component to storybook

DeLettenhove:
Add VideoBox component to storybook.

### DIFF
--- a/src/stories/VideoBox.stories.js
+++ b/src/stories/VideoBox.stories.js
@@ -1,0 +1,21 @@
+import VideoBox from '~/components/video-box/VideoBox'
+import videoImg from '~/assets/img/guest-home-page/videoImg.png'
+
+export default {
+  title: 'VideoBox',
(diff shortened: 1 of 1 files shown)


src/stories/VideoBox.stories.js
```diff
@@ -0,0 +1,33 @@
...
+      description: 'Variable of the image'
+    }
+  }
+}
+
+const Template = (args) => <VideoBox {...args} />
+
+export const Desktop = Template.bind({})
+
+Desktop.args = {
```
rostyslaw27, line 19: leave only one state because props are the same everywhere

rostyslaw27 (changes requested):

src/stories/VideoBox.stories.js
DeLettenhove: Fixed.

rostyslaw27 (approved):

PR #281, opened 2022-08-29
fixed header overlap content

greentoss:
fixed overlap content under header

### DIFF
--- a/src/containers/app-content/AppContent.js
+++ b/src/containers/app-content/AppContent.js
@@ -11,7 +11,7 @@ const styles = {
     height: '100vh',
     '.MuiToolbar-root + div': {
       flexGrow: 1,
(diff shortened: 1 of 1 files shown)


Yevhendzyuba (approved):

src/containers/app-content/AppContent.js
```diff
@@ -10,7 +10,12 @@ const styles = {
     flexDirection: 'column',
     height: '100vh',
     '.MuiToolbar-root + div': {
-      flexGrow: 1
+      flexGrow: 1,
+      marginTop: {
```
rostyslaw27, line 14: you can rename it from marginTop to mt

greentoss: renamed

PR #421, opened 2023-01-31
fixed signup form validation

greentoss:
- fields are red if error on blur also as on change
- submit button disabled untill all form fields are valid and filled
- checkbox is active untill all form fields are valid and filled
bug3

### DIFF
--- a/src/containers/guest-home-page/signup-form/SignupForm.js
+++ b/src/containers/guest-home-page/signup-form/SignupForm.js
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react'
+import { useState, useMemo, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 import HashLink from '~/components/hash-link/HashLink'
 import { useSelector } from 'react-redux'
@@ -18,7 +18,6 @@ const SignupForm = ({ handleSubmit, handleChange, handleBlur, data, errors, clos
   const { t } = useTranslation()
   const { privacyPolicy, termOfUse } = guestRoutes
   const [buttonDisabled, setButtonDisabled] = useState(true)
-  const [checkDisabled, setCheckDisabled] = useState(true)
   const { inputVisibility: passwordVisibility, showInputText: showPassword } = useInputVisibility(errors.password)
   const { inputVisibility: confirmPasswordVisibility, showInputText: showConfirmPassword } = useInputVisibility(
     errors.confirmPassword
@@ -29,13 +28,11 @@ const SignupForm = ({ handleSubmit, handleChange, handleBlur, data, errors, clos
     setButtonDisabled(!buttonDisabled)
   }
 
-  const handleIsFormValid = () => {
-    return Object.values(errors).every((elem) => elem === undefined) && Object.values(data).every((elem) => elem !== '')
-  }
+  const hasErrors = useCallback(() => Object.values(errors).some((elem) => elem === undefined), [errors])
 
-  useEffect(() => {
-    handleIsFormValid() ? setCheckDisabled(false) : setCheckDisabled(true)
-  }, [errors, handleIsFormValid])
+  const emptyData = useCallback(() => Object.values(data).every((elem) => elem !== ''), [data])
+
+  const isValid = useMemo(() => hasErrors() && emptyData(), [emptyData, hasErrors])
 
   const policyAgreement = (
     <Box sx={ styles.box }>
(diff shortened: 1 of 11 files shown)


KhrystynaPavlikovska (approved):

src/containers/guest-home-page/signup-form/SignupForm.js
```diff
@@ -28,6 +29,14 @@ const SignupForm = ({ handleSubmit, handleChange, handleBlur, data, errors, clos
     setButtonDisabled(!buttonDisabled)
   }
```
maarekua, line 31: const isValid = useMemo(() => hasErros || noData, [hasErrors, noData])

greentoss: remade through using useMemo. the result is the same as in gif before
once again - we check errors object for undefined values 
and check data object for not empty strings
Screenshot_2
so i used .some() for errors object, because its elements added only if error ocurs
and used .every() for data object, because it has 5 elements and every of them has to be not undefined

another variants are cousing unappropriate behavior, and enable user to submit empty form

maarekua: Still we can use the useMemo hook.

src/containers/guest-home-page/signup-form/SignupForm.js
```diff
@@ -28,6 +28,12 @@ const SignupForm = ({ handleSubmit, handleChange, handleBlur, data, errors, clos
...
 
+  const hasErrors = useCallback(() => Object.values(errors).some((elem) => elem === undefined), [errors])
+
+  const emptyData = useCallback(() => Object.values(data).every((elem) => elem !== ''), [data])
```
MickhaelSh, line 33: if u need only result and not the function use useMemo instead useCallback

greentoss: yes, i used useCallback in those separete functions. now i refactored, removed separete funcs and putt everything into useMemo

src/containers/guest-home-page/signup-form/SignupForm.js
greentoss: yes i had used it, refactored a bit

src/utils/validations/login.js
```diff
@@ -31,7 +31,7 @@ export const lastName = (value) => {
 
 export const confirmPassword = (password, data) => {
   let helperText
-  if (password !== data.confirmPassword) {
+  if (password !== data.password) {
     helperText = 'common.errorMessages.passwordsDontMatch'
```
maarekua, line 35: helperText = emptyField(password, password !== data.password ? 'common.errorMessages.passwordsDontMatch' : '')

greentoss: refactored using ternar
now helpertext is a string anyway, either full or empty, 
so i had to change it to be string in all validations.

src/containers/guest-home-page/signup-form/SignupForm.js
```diff
@@ -131,6 +136,7 @@ const SignupForm = ({ handleSubmit, handleChange, handleBlur, data, errors, clos
       <Box sx={ styles.checkboxContainer }>
         <FormControlLabel
           control={ <Checkbox /> }
+          disabled={ isValid ? false : true }
```
MickhaelSh, line 139: u can use just isValid if this is boolean, or Boolean(isValid), ternary is not needed

src/utils/validations/login.js
```diff
@@ -31,9 +31,6 @@ export const lastName = (value) => {
 
 export const confirmPassword = (password, data) => {
   let helperText
-  if (password !== data.confirmPassword) {
-    helperText = 'common.errorMessages.passwordsDontMatch'
-  }
-  helperText = emptyField(password, helperText)
+  helperText = emptyField(password, password !== data.password ? 'common.errorMessages.passwordsDontMatch' : '')
```
MickhaelSh, line 34: what is this? omg

MickhaelSh: delete helper text and return this ternary, or with using if operator use few returns, 
if (password !== data.confirmPassword){
return error)
return result

greentoss: returned to previous variant but mentioninh helpertext as string

src/containers/guest-home-page/signup-form/SignupForm.js
greentoss: fixed

PR #503, opened 2023-03-02
472/Implemented offers view switcher

mxrcury:
Implement a switcher that can change view of offers (horizontal and square versions):

[FIGMA](https://www.figma.com/file/liJZDYhUnDP15Pi9bipDcv/Space2Study-(Students)?node-id=3193%3A59961&t=RQ8FuzVvyCQM7MdM-1)
image
hovered
Screenshot from 2023-03-03 07-42-56

### DIFF
--- a/src/components/view-switcher/ViewSwitcher.jsx
+++ b/src/components/view-switcher/ViewSwitcher.jsx
@@ -5,7 +5,7 @@ import FormatListBulletedIcon from '@mui/icons-material/FormatListBulleted'
 import GridViewIcon from '@mui/icons-material/GridView'
 
 import { cardsViews } from '~/constants'
-import { styles } from './ViewSwitcher.styles'
+import { styles } from '~/components/view-switcher/ViewSwitcher.styles'
 
 const ViewSwitcher = ({ setOffersView }) => {
   const [cardsView, setCardsView ] = useState(cardsViews.inline)
@@ -28,7 +28,8 @@ const ViewSwitcher = ({ setOffersView }) => {
       <ToggleButton
         aria-label='grid card view'
         onClick={ changeCardsView }
-        selected={ cardsView === cardsViews.grid } sx={ { ...styles.toggleButton } }
+        selected={ cardsView === cardsViews.grid }
+        sx={ styles.toggleButton }
         value={ cardsViews.grid }
       >
         <GridViewIcon sx={ styles.icon }  />
--- a/src/components/view-switcher/ViewSwitcher.jsx
+++ b/src/components/view-switcher/ViewSwitcher.jsx
@@ -1,4 +1,3 @@
-import { useState } from 'react'
 import ToggleButton from '@mui/material/ToggleButton'
 
 import FormatListBulletedIcon from '@mui/icons-material/FormatListBulleted'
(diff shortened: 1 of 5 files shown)


MickhaelSh (commented):
Add description to pr, and screenshot of your work

src/components/view-switcher/ViewSwitcher.jsx
```diff
@@ -0,0 +1,40 @@
...
+import GridViewIcon from '@mui/icons-material/GridView'
+
+import { cardsViews } from '~/constants'
+import { styles } from './ViewSwitcher.styles'
+
+const ViewSwitcher = ({ setOffersView }) => {
+  const [cardsView, setCardsView ] = useState(cardsViews.inline)
+
+  const changeCardsView = (_event,view) => {
+    setOffersView(view)
```
MickhaelSh, line 14: do you need second state where you set the same value?

src/components/view-switcher/ViewSwitcher.jsx
```diff
@@ -0,0 +1,40 @@
+import { useState } from 'react'
+import ToggleButton from '@mui/material/ToggleButton'
+
+import FormatListBulletedIcon from '@mui/icons-material/FormatListBulleted'
+import GridViewIcon from '@mui/icons-material/GridView'
+
+import { cardsViews } from '~/constants'
+import { styles } from './ViewSwitcher.styles'
```
MickhaelSh, line 8: use ~ for imports

src/components/view-switcher/ViewSwitcher.jsx
```diff
@@ -0,0 +1,40 @@
...
+  const changeCardsView = (_event,view) => {
+    setOffersView(view)
+    setCardsView(view)
+  }
+
+  return (
+    <>
+      <ToggleButton
+        aria-label='inline card view' onClick={ changeCardsView }
+        selected={ cardsView === cardsViews.inline }
```
MickhaelSh, line 22: move

src/components/view-switcher/ViewSwitcher.jsx
```diff
@@ -0,0 +1,40 @@
...
+        sx={ { ...styles.toggleButton, marginRight:'8px' } }
+        value={ cardsViews.inline }
+      >
+        <FormatListBulletedIcon sx={ styles.icon } />
+      </ToggleButton>
+      <ToggleButton
+        aria-label='grid card view'
+        onClick={ changeCardsView }
+        selected={ cardsView === cardsViews.grid } sx={ { ...styles.toggleButton } }
```
MickhaelSh, line 31: no need for destructuring in sx

src/components/view-switcher/ViewSwitcher.jsx
```diff
@@ -0,0 +1,40 @@
...
+    setOffersView(view)
+    setCardsView(view)
+  }
+
+  return (
+    <>
+      <ToggleButton
+        aria-label='inline card view' onClick={ changeCardsView }
+        selected={ cardsView === cardsViews.inline }
+        sx={ { ...styles.toggleButton, marginRight:'8px' } }
```
MickhaelSh, line 23: you can add this in styles file

mxrcury: in this way style "marginRight:'8px'" will be in all buttons but i need only in one

src/components/view-switcher/ViewSwitcher.jsx
mxrcury: useState inside is for setting value of current view to see active button
you mean it would be better if state of current view will come
from props?

src/components/view-switcher/ViewSwitcher.jsx
mxrcury: fixed

src/components/view-switcher/ViewSwitcher.jsx
mxrcury: fixed

src/components/view-switcher/ViewSwitcher.jsx
MickhaelSh: if you have a state that controls the view setting outside you don't need it here

src/components/view-switcher/ViewSwitcher.jsx
MickhaelSh: you can use variables and destructure in styles file , const toggleButton = {some styles} then in styles object use grid: toggleButton , inline : {...toggleButton, mr: '8px'}

mxrcury: refactored it

src/components/view-switcher/ViewSwitcher.jsx
mxrcury: i have deleted useState and change useing local state to state from props

mxrcury: @MickhaelSh , would it be a good idea if i will create separate file 'data.js' that will be containing array of objects with field needed for every "ToggleButton" props and then just map them or it will be redundant?

MickhaelSh (approved):

src/components/view-switcher/ViewSwitcher.jsx
```diff
@@ -0,0 +1,37 @@
...
+import Box from '@mui/material/Box'
+
+import FormatListBulletedIcon from '@mui/icons-material/FormatListBulleted'
+import GridViewIcon from '@mui/icons-material/GridView'
+
+import { cardsViews } from '~/constants'
+import { styles } from '~/components/view-switcher/ViewSwitcher.styles'
+
+const ViewSwitcher = ({ setOffersView, offersView }) => {
+  const changeOffersView = (_event, view) => setOffersView(view)
```
maarekua, line 11: It should emit to parent and not set state...

maarekua (changes requested):

src/components/view-switcher/ViewSwitcher.jsx
mxrcury: you mean it has to be incoming function that arguments will take not only one view but also spreading incoming other filters or what?  @maarekua

maarekua (approved):

Tolik170 (approved):

PR #522, opened 2023-03-07
add routes for pages

Yevhendzyuba:


Tolik170 (approved):

src/router/constants/guestRoutes.js
```diff
@@ -11,5 +11,8 @@ export const guestRoutes = {
     whatCanYouDo: { route: 'whatCanYouDo', path: '/#whatCanYouDo' },
     howItWorks: { route: 'howItWorks', path: '/#howItWorks' },
     whoWeAre: { route: 'whoWeAre', path: '/#whoWeAre' }
-  }
+  },
+  categories: { route: 'categories', path: '/categories' },
```
maarekua, line 15: These routes should be nested. Lets say if user picked lanuages, english...

languages/english/offers

if he pick se all offers...

categories/subjects/offers..

not exactly but something like that. We can discuss it. But it should be totally separate. Think about a suggestions.

maarekua (approved):

src/router/constants/guestRoutes.js
Yevhendzyuba: fixed

maarekua (approved):

PR #536, opened 2023-03-15
Feature/534/checkbox list component

mxrcury:
image


src/stories/CheckboxList.stories.jsx
```diff
@@ -0,0 +1,30 @@
+import CheckboxList from '~/components/checkbox-list/CheckboxList'
+
+export default {
+  title: 'CheckboxList',
+  component: CheckboxList,
+  argTypes: {
+    items: [],
+    title: {
+      type: 'string',
```
MickhaelSh, line 9: use ts and import interface from your component so you don't need write type manualy

mxrcury: thx but as i know we dont need to write stories in ts

MickhaelSh: it makes stories more informative, so it will be better if we will use it

mxrcury: agree with you

src/components/app-select/AppSelect.tsx
```diff
@@ -1,24 +1,32 @@
...
 
-const AppSelect = ({ setValue, value, fields, selectTitle = '' }) => {
+interface AppSelectProps {
+  setValue:(value:string) => void
+  value:string,
+  fields:SelectField[],
+  selectTitle?:string,
+}
+
+const AppSelect = ({ setValue, value, fields, selectTitle = '' }:AppSelectProps) => {
```
MickhaelSh, line 18: do as in CheckboxList with FC

src/components/app-select/AppSelect.tsx
```diff
@@ -1,24 +1,32 @@
...
   const fieldsList = fields.map(field => (
     <MenuItem key={ field.value } value={ field.value }>
       { t(field.title) }
     </MenuItem>)
   )
   const titleEl = selectTitle.length ? (
-    <Typography sx={ styles.selectTitle } variant='subtitle1' >
+    <Typography aria-label='select-title'  sx={ styles.selectTitle } variant='subtitle1' >
       { t(selectTitle) }
       :
```
MickhaelSh, line 31: move this to translation consts

mxrcury: aria-label?

src/components/app-select/AppSelect.tsx
mxrcury: ok, changed

src/components/app-select/AppSelect.tsx
MickhaelSh: ':'

maarekua (approved):

maarekua (approved):

src/components/app-select/AppSelect.tsx
mxrcury: fixed

PR #537, opened 2023-03-15
Added price filter component

MickhaelSh:
Added price filter component
image
image


src/hooks/use-debounce.tsx
```diff
@@ -0,0 +1,25 @@
+import { useRef, useEffect } from 'react'
+
+type TimerType = ReturnType<typeof setTimeout>;
+type CallbackType<T> = (...args: T[]) => void;
+
+export const useDebounce = <T,>(callback: CallbackType<T>, delay = 500): CallbackType<T> => {
```
Yurenko, line 6: Typo, delete comma

MickhaelSh: I can`t because if I delete vite thinks that is jsx and throws an error, it can be fixed by changing the extension to ts from tsx but then vite can`t compile it. We need to fix vite config in other issue

Yurenko: Ok. I gotcha, so now we have a new issue.

src/types/components/price-filter/price-filter.types.ts
```diff
@@ -0,0 +1,20 @@
+export type InputPrice = number|null
+export type PriceArray = [number, number]
+export type InputPriceArray = [InputPrice, InputPrice]
+
+export interface CheckInputChange {
+    inputValue: InputPrice
```
Yurenko, line 6: Maybe need to separate type and interface in different files

src/components/price-filter/PriceFilter.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const prices = value as PriceArray
+    
+    setPrices(prices)
+    debouncedOnChange(prices)
+  }
+
+  const handleInputChange = ({ target }: React.ChangeEvent<HTMLInputElement>) => {
+    const inputIndex  = Number(target.id)
+    const inputValue = target.value ? Number(target.value) : null
+    const notValidInput = checkPriceInput(inputValue)
```
maarekua, line 46: if(checkPriceInput(inputValue)) {
const newPrices = createNewState({ prices, inputValue, inputIndex })
    setPrices(newPrices)
    debouncedOnChange(newPrices)
}

MickhaelSh: fixed

PR #572, opened 2023-03-22
implement categories with search  block

Tolik170:
image


src/components/search-with-filters/SearchWithFilters.styles.js
```diff
@@ -0,0 +1,27 @@
+export const styles = {
+  container: {
+    backgroundColor: 'basic.white',
+    mb: '50px',
+    p: '30px 45px',
+    borderRadius: '70px',
+    display: 'flex',
+    justifyContent: 'space-between',
+    alignItems: 'center'
```
Yurenko, line 9: Add box-shadow to this component

src/components/app-auto-complete/AppAutoComplete.jsx
```diff
@@ -12,38 +12,45 @@ const defaultFilterOptions = (options, state) => {
 
 const AppAutoComplete = ({
   disabled = false,
+  freeSolo = false,
   fieldValue,
   filterOptions = defaultFilterOptions,
+  ListboxProps = { style: { maxHeight: 150 } },
   loading = false,
   onChange,
+  onInputChange,
```
MickhaelSh, line 21: u dont need onChange and onInputChange here also freeSolo and disabled it will be passed as ...props, because u just pass it without modifying

Tolik170: I use props for TextField inside AutoComplete, not for itself, so when i pass disable, it will disable only input but i can click on autocomplete and see list of options

freeSolo is autocomplete prop not TextField

OnChange set value only if i chose it from list of options
onInputChange set value when i choose it from list or write in textfield, they can be used for different cases, so i decided to leave both(for example in choosing cities or countries, the better way is to chose only from options list)

maarekua (approved):

PR #574, opened 2023-03-23
Added popular categories

Yurenko:
Created popular categories block.
Also added adaptivity.

image


src/components/popular-categories/PopularCategories.tsx
```diff
@@ -0,0 +1,80 @@
...
+  items: CategoryInterface[]
+  loading: boolean
+}
+
+const PopularCategories: FC<PopularCategoriesProps> = ({ title, description, items, loading }) => {
+  const { t } = useTranslation()
+  const navigate = useNavigate()
+  const { isDesktop, isTablet, isMobile } = useBreakpoints()
+
+  const itemsToShow = (isDesktop && 9) || (isTablet && 6) || (isMobile && 4) || 9
```
Tolik170, line 29: Do you need || 9 in the end?

Yurenko: I think yes because itemsToShow can return "false | 9 | 6 | 4". And we can't pass false to the method slice. I decided to pass the default value. May you know a better solution?

Tolik170: i guess it can`t be false:
xs - sm  = mobile
sm - md = tablet
md and upper = desktop

MickhaelSh (approved):

src/components/popular-categories/PopularCategories.tsx
Yurenko: I also think so.
You can see what happens when I leave without || 9
image

Tolik170: ok

Tolik170 (approved):

src/components/popular-categories/PopularCategories.tsx
```diff
@@ -0,0 +1,80 @@
...
+      <TitleWithDescription
+        description={ description }
+        descriptionStyles={ styles.descriptionStyles }
+        title={ title }
+        titleStyles={ styles.titleStyles }
+      />
+      { loading && !items ? (
+        <Loader size={ 70 } />
+      ) : (
+        <ClickableCardList
```
maarekua, line 69: Router Link

maarekua (approved):

maarekua (approved):

PR #575, opened 2023-03-24
Feature/567/Implemented filter bar menu

mxrcury:
- Combined all elements together for top filter bar menu on Find Offers page

### Desktop
Screenshot from 2023-03-24 12-57-16
### Tablet:
Screenshot from 2023-03-24 12-57-51
### Mobile
Screenshot from 2023-03-24 12-58-21
How looks data from getting filters:
Screenshot from 2023-03-24 13-00-27
- Also added filter title component
Screenshot from 2023-03-24 12-55-33
- Will make unit test for this container some time later cause at this moment there is some troubles with running all tests


src/components/filters-title/FiltersTitle.styles.ts
```diff
@@ -0,0 +1,31 @@
...
+  },
+  chosenFiltersQty :{
+    borderRadius: '50%',
+    width: '24px',
+    height: '25px',
+    backgroundColor: 'primary.400',
+    color:'white',
+    fontWeight:'500',
+    textAlign: 'center',
+    fontSize: '14px',
```
MickhaelSh, line 28: fontsize fontweight can be replaced as typography: 'subtitle2'

src/components/filters-title/FiltersTitle.tsx
```diff
@@ -0,0 +1,33 @@
...
+interface FiltersTitleProps {
+    chosenFiltersQty?:number,
+    handleOpenFilters: () => void,
+}
+
+const FiltersTitle:FC<FiltersTitleProps> = ({ chosenFiltersQty = 0, handleOpenFilters }) => {
+  const { t } = useTranslation()
+  
+  return (
+    <Box sx={ { ...styles.container } } >
```
MickhaelSh, line 19: no need for destructuring

src/components/filters-title/FiltersTitle.tsx
```diff
@@ -0,0 +1,33 @@
...
+      <Box onClick={ handleOpenFilters } sx={ styles.cursorContainer } >
+        <FilterListIcon sx={ styles.icon } />
+        <Typography sx={ styles.title } variant='h6' >
+          { t('filters.filtersListTitle') }
+        </Typography>
+      </Box>
+      <Typography sx={ styles.chosenFiltersQty } >
```
MickhaelSh, line 26: I think it should be hidden when the quantity is 0

src/containers/find-offer/filter-bar-menu/FilterBarMenu.constants.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+  
+export const sortByFields = [
+  {
+    value: 'popularity',
+    title: 'filters.sortBy.sortByFields.popularity'
+  },
+  {
+    value: 'newest',
```
MickhaelSh, line 21: those values needed for sorting offers on backEnd check how it should be named

src/containers/find-offer/filter-bar-menu/FilterBarMenu.styles.ts
```diff
@@ -0,0 +1,13 @@
+const container = {
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'space-between'
+}
+
+export const styles = {
+  container,
+  rightContainer:{ ...container },
```
MickhaelSh, line 9: it`s redundant

src/containers/find-offer/filter-bar-menu/FilterBarMenu.constants.ts
```diff
@@ -0,0 +1,38 @@
+import { CardsViewEnums } from '~/types/findOffers/enums/viewSwitcher.enums'
+
+
+export const switcherOptions = {
```
MickhaelSh, line 4: is it needed ? you are using the same object in component

src/containers/find-offer/filter-bar-menu/FilterBarMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+  
+
+  return (
+    <Box sx={ isMobile ? styles.mobileContainer : styles.container } >
+      <FiltersTitle chosenFiltersQty={ chosenFiltersQty } handleOpenFilters={ openFilters } />
+      { isDesktop ? (
+        <AppContentSwitcher
+          active={ barMenuFilters.isActiveOffersType }
+          handleChange={ handleOffersType }
+          styles={ {} }
```
MickhaelSh, line 81: remove

src/containers/find-offer/filter-bar-menu/FilterBarMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+
+interface FilterBarMenuProps {
+  chosenFiltersQty:number,
+  openFilters: () => void,
+  getFilters: (filters: BarMenuFilters) => void
+}
+
+const FilterBarMenu: FC<FilterBarMenuProps> = ({ chosenFiltersQty, openFilters, getFilters }) => {
+  const [barMenuFilters, setBarMenuFilters] = useState<BarMenuFilters>(initialBarMenuFilters)
```
MickhaelSh, line 26: this state should be outside so we can control cards with it

src/containers/find-offer/filter-bar-menu/FilterBarMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const handleSortBy = (sortBy: string) => {
+    setBarMenuFilters((filters) => {
+      const updatedFilters = { ...filters, sortBy }
+      getFilters(updatedFilters)
+
+      return updatedFilters
+    })
+  }
+
+  const handleCardView = (view: CardsViewTypes) => {
```
MickhaelSh, line 64: card view should be state outside so we can control cards view with it , don't pass it to filters

src/containers/find-offer/filter-bar-menu/FilterBarMenu.constants.ts
mxrcury: as for me it was looking cleaner, but ok, i have refactored

src/containers/find-offer/filter-bar-menu/FilterBarMenu.constants.ts
mxrcury: does we have sorting service of offers on Back End?

src/containers/find-offer/filter-bar-menu/FilterBarMenu.styles.ts
mxrcury: i have refactored

src/containers/find-offer/filter-bar-menu/FilterBarMenu.tsx
mxrcury: card view will be incoming from props of this component?

src/containers/find-offer/filter-bar-menu/FilterBarMenu.tsx
mxrcury: i cant, cause there is no default value inside of this component so prop styles is necessary...

src/containers/find-offer/filter-bar-menu/FilterBarMenu.constants.ts
MickhaelSh: values should be named as fields in model

src/containers/find-offer/filter-bar-menu/FilterBarMenu.tsx
MickhaelSh: so add default it will be much better

src/containers/find-offer/filter-bar-menu/FilterBarMenu.tsx
MickhaelSh: yes

src/containers/find-offer/filter-bar-menu/FilterBarMenu.tsx
```diff
@@ -0,0 +1,94 @@
...
+interface FilterBarMenuProps {
+  chosenFiltersQty:number,
+  openFilters: () => void,
+  getFilters: (filters: BarMenuFilters) => void,
+  handleOffersView:(view: CardsViewTypes) => void,
+  offersView: CardsViewTypes
+}
+
+const FilterBarMenu: FC<FilterBarMenuProps> = ({ chosenFiltersQty, openFilters, getFilters, handleOffersView, offersView }) => {
+  const [barMenuFilters, setBarMenuFilters] = useState<BarMenuFilters>(initialBarMenuFilters)
```
MickhaelSh, line 28: i dont think we need state here, we will manage all filters outside, just pass them as prop

mxrcury: but maybe passing all filters will be unnecessary for this component, so it will be better to have filters only it needs locally and on changing just move them outside and setting them in state of others filters

MickhaelSh: if you don't pass all filters you would not know how many active filters you have to chow on the icon near "Filter"

MickhaelSh: your filters and filters from sidebar will be in one object

mxrcury: this quantity we can count outside and just pass a number of it, there is such prop "chosenFiltersQty"

MickhaelSh: yep but your filters affect on cards show also as other filters I don't see the point of why they have to be separate

mxrcury: i think your way is good, but how do you think is it good that this container will be getting from props all filters, that it needs and that does not need

MickhaelSh: for me, it`s a better solution than one more state which can be avoided

mxrcury: okey, thx

MickhaelSh (approved):

Yevhendzyuba (approved):

PR #582, opened 2023-03-28
Combined all components into category page

tsivkadi:
Combined all components into category page + fixed bag in translation files

### DIFF
--- a/src/constants/translations/en/index.js
+++ b/src/constants/translations/en/index.js
@@ -23,7 +23,7 @@ import completeProfile from './complete-profile.json'
 import tutorProfilePage from './tutor-profile-page.json'
 import adminTable from './admin-table.json'
 import constant from './constant.json'
-import findOffers from './find-offer-page.json'
+import findOffer from './find-offer-page.json'
 import filters from './filters.json'
 
 const en = {
@@ -53,7 +53,7 @@ const en = {
     tutorProfilePage,
     adminTable,
     constant,
-    findOffers,
+    findOffer,
     filters
   }
 }
--- a/src/pages/categories/Categories.jsx
+++ b/src/pages/categories/Categories.jsx
@@ -2,11 +2,13 @@ import Container from '@mui/material/Container'
 import Box from '@mui/material/Box'
 
 import ExploreCategories from '~/containers/categories-page/explore-categories/ExploreCategories'
+import OfferRequestBlock from '~/containers/find-offer/OfferRequestBlock'
 
 const Categories = () => {
   return (
     <Box sx={ { backgroundColor: 'backgroundColor', flex: 1 } }>
-      <Container>
+      <Container sx={ { pt: 6 } } >
+        <OfferRequestBlock />
         <ExploreCategories />
       </Container>
     </Box>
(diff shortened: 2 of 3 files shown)


KhrystynaPavlikovska (approved):

src/constants/translations/en/index.js
```diff
@@ -23,7 +23,7 @@ import completeProfile from './complete-profile.json'
 import tutorProfilePage from './tutor-profile-page.json'
 import adminTable from './admin-table.json'
 import constant from './constant.json'
-import findOffers from './find-offer-page.json'
+import findOffer from './find-offer-page.json'
 import filters from './filters.json'
 
 const en = {
```
Yurenko, line 29: Are you sure that you fixed an error in the correct place? 
I think it is wrong because you fixed only your problem but broke down in another place.
Check it, please)
Try will fix here your problem and don't change the name of this import "findOffers"
image

src/pages/categories/Categories.jsx
```diff
@@ -2,11 +2,13 @@ import Container from '@mui/material/Container'
 import Box from '@mui/material/Box'
 
 import ExploreCategories from '~/containers/categories-page/explore-categories/ExploreCategories'
+import OfferRequestBlock from '~/containers/find-offer/OfferRequestBlock'
 
 const Categories = () => {
   return (
     <Box sx={ { backgroundColor: 'backgroundColor', flex: 1 } }>
-      <Container>
+      <Container sx={ { pt: 6 } } >
```
MickhaelSh, line 10: do container as top component instead of box, and move sx to syles file

src/constants/translations/en/index.js
```diff
@@ -23,7 +23,7 @@ import completeProfile from './complete-profile.json'
 import tutorProfilePage from './tutor-profile-page.json'
 import adminTable from './admin-table.json'
 import constant from './constant.json'
-import findOffers from './find-offer-page.json'
+import findOffer from './find-offer-page.json'
```
MickhaelSh, line 26: why do you rename it? if this page is find-offers across the app, if we need to rename we should rename all files

Yurenko: You are right. I attached the screen where needs to fix his problem)

src/pages/categories/Categories.jsx
Yevhendzyuba: @MickhaelSh  I think there are no need to move one styles to file

src/constants/translations/en/index.js
tsivkadi: fixed

src/pages/categories/Categories.jsx
MickhaelSh: for all make sense, if it is styles it should be in styles file, not everywhere

MickhaelSh (approved):

Yurenko (approved):

src/pages/categories/Categories.jsx
tsivkadi: fixed

Yevhendzyuba (approved):

maarekua (approved):

PR #610, opened 2023-04-21
Converted AppPagination to TS

mxrcury:
Converted component AppPagination to TypeScript


src/types/common/enums/common.enums.ts
```diff
@@ -6,3 +6,9 @@ export enum ProficiencyLevelEnums {
   Professional = 'Professional',
   Specialized = 'Specialized'
 }
+
+export enum Size {
```
MickhaelSh, line 10: we decided that in enums we use Enum keyword in the end of the name

src/components/app-pagination/AppPagination.tsx
```diff
@@ -1,17 +1,29 @@
...
+  pageSize: number
+  setCurrentPage: (page: number) => void
+}
+
+const AppPagination: FC<AppPaginationProps> = ({
+  size = Size.Medium,
   page,
   itemsCount,
   pageSize,
   setCurrentPage
```
MickhaelSh, line 22: onChange instead of setCurrentPage. also add posibility to pass additional props to pagination, we don`t need to specify page in props it will go in additional props

MickhaelSh (commented):
Always provide description to PR

src/components/app-pagination/AppPagination.tsx
```diff
@@ -1,17 +1,29 @@
...
+  itemsCount: number
+  pageSize: number
+  setCurrentPage: (page: number) => void
+}
+
+const AppPagination: FC<AppPaginationProps> = ({
+  size = Size.Medium,
   page,
   itemsCount,
   pageSize,
```
MickhaelSh, line 21: also pageSize is not cleat, itemsPerPage will be better

src/types/common/enums/common.enums.ts
mxrcury: okey, changed

src/components/app-pagination/AppPagination.tsx
mxrcury: changed

maarekua (approved):

src/components/app-pagination/AppPagination.tsx
mxrcury: Added additional props and changed onChange function naming

MickhaelSh (approved):

PR #658, opened 2023-04-28
Accordion for FAQ

mxrcury:
Created component for accordion for FAQ on Offer Details page, also i have added to storybook and wrote a unit test
Component with title:
| With Title | Without title |
|--------|--------|
| ready-component | ready-component-no-title |


src/components/accordion/Accordion.styles.ts
```diff
@@ -37,7 +37,9 @@ export const styles = {
       '& h6': {
         color: 'basic.white'
       }
-    }
+    },
+    summary: {},
+    details: {}
```
Yurenko, line 42: If you don't use these styles, delete them

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+        return prevActiveQuestions.filter(
+          (prevActiveQuestion) => prevActiveQuestion !== activeQuestion
+        )
+      } else {
+        return [...prevActiveQuestions, activeQuestion]
+      }
+    })
+  }
+
+  const { t } = useTranslation()
```
Yurenko, line 40: Move it above

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <AccordionSummary
+              sx={styles.accordion.summary}
+              data-testid={`accordion-summary-${index}`}
+              expandIcon={
+                showIcon ? (
+                  <ArrowForwardIosSharpIcon
+                    sx={styles.icon}
+                    data-testid='accordion-icon'
+                  />
+                ) : null
```
Yurenko, line 67: Move out from jsx

src/components/accordion/Accordion.styles.ts
mxrcury: they are needed for TS

src/components/questions-accordion/QuestionsAccordion.tsx
mxrcury: okey

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+      }
+    })
+  }
+
+  const { t } = useTranslation()
+
+  return (
+    <Box sx={title ? styles.container : {}}>
+      {title ? (
+        <Typography sx={styles.title} variant='h5'>
```
Yurenko, line 45: We use TS, add to VariantEnum h5 and use it.
variant={VariantEnum .h5}

src/components/questions-accordion/QuestionsAccordion.tsx
mxrcury: you mean to move full list of items to a variable?

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+                  <ArrowForwardIosSharpIcon
+                    sx={styles.icon}
+                    data-testid='accordion-icon'
+                  />
+                ) : null
+              }
+            >
+              <Typography
+                sx={styles.accordion.title(activeQuestions.includes(index))}
+                variant={'h6'}
```
Yurenko, line 72: Here the same

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+              >
+                {t(item.title)}
+              </Typography>
+            </AccordionSummary>
+            <AccordionDetails>
+              <Typography
+                sx={styles.accordion.caption(showIcon)}
+                data-testid={`accordion-description-${index}`}
+                variant={'body2'}
```
Yurenko, line 81: And here

src/components/questions-accordion/QuestionsAccordion.tsx
mxrcury: yeap, i know it but i don't see such variant in enums

Yurenko: You can add this variant

src/components/questions-accordion/QuestionsAccordion.tsx
Yurenko: Yeap

src/components/questions-accordion/QuestionsAccordion.tsx
mxrcury: okey 
firstly, I wanted to add to this enums but i thought that this enums are only for variants similar to existing but not for text

src/components/accordion/Accordion.styles.ts
MickhaelSh: so do them not required in TS

src/components/questions-accordion/QuestionsAccordion.tsx
mxrcury: okey

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { t } = useTranslation()
+
+  return (
+    <Box sx={title ? styles.container : {}}>
+      {title ? (
+        <Typography sx={styles.title} variant='h5'>
+          {t(title)}
+        </Typography>
+      ) : null}
+      {items.map((item, index) => (
```
MickhaelSh, line 49: you are using here code almost similar to Accordions component, why you couldn't use it here?

mxrcury: because Accordions component is not what is need to create this QuestionsAccordion for the next reasons:
- QuestionsAccordion component

mxrcury: because Accordions component is not what is need to create this QuestionsAccordion for the next reasons:
- QuestionsAccordion component

mxrcury: because Accordions component is not what i needed to create QuestionsAccordion for the next reasons:
- QuestionsAccordion is using logic not the same as Accordions, cause in the first one we can open all items as it is shown in Figma mark up, in the Accordion we can open only and if we open one other will be close automatically
- QuestionsAccordion component is using another icon than Accordions
- Accordions had some props for MUI Accordion component that were no need in QuestionsAccordion so i thought it will be too much props to add to Accordions and it will be cleaner and better to create from MUI Accordion component
- Also some micro moments connected with styling, where i couldn't make looking of existing Accordion component to what i needed for QuestionsAccordion, mostly it was concerned with positioning

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+        return [...prevActiveQuestions, activeQuestion]
+      }
+    })
+  }
+
+  const { t } = useTranslation()
+
+  return (
+    <Box sx={title ? styles.container : {}}>
+      {title ? (
```
Tolik170, line 44: you can move this outside return and check with &&, so you will not need to use : {} in the end:

const variableName = title && ( 'what you need to render' )

src/components/questions-accordion/QuestionsAccordion.tsx
Tolik170, line 49: same as above, move this map outdide to variable

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+            disableGutters
+            onChange={() => onChange(index)}
+            elevation={0}
+            style={styles.accordion.root}
+          >
+            <AccordionSummary
+              sx={styles.accordion.summary}
+              data-testid={`accordion-summary-${index}`}
+              expandIcon={
+                showIcon ? (
```
Tolik170, line 62: use && instead of ternary

src/components/accordion/Accordion.styles.ts
mxrcury: @MickhaelSh I have tried by removing summary from styles and making condition with &&, where sx={accordionStyle.summary && accordionStyle.summary} but TS still need to have these fields

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,93 @@
...
+      } else {
+        return [...prevActiveQuestions, activeQuestion]
+      }
+    })
+  }
+
+  const { t } = useTranslation()
+
+  return (
+    <Box sx={title ? styles.container : {}}>
```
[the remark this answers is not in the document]
mxrcury: in this way with TS, sx prop needs something but not empty prop, so with && it doesn't work well

src/components/questions-accordion/QuestionsAccordion.tsx
MickhaelSh: for me its looking like minor changes, we should not create new component for ewery need we have

MickhaelSh: if accordions component is too different you have to create a more general component which we could reuse, but yours is too specified, and when we will need another accordion we will create one more what is bad

mxrcury: i understand that reusing is much better and needs performing by maximum, but i still don't know how combine two different logis of opening accordions, depends on what it should change, do you have any ideas?

src/components/questions-accordion/QuestionsAccordion.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { AccordionItem, VariantEnum } from '~/types'
+import { styles } from '~/components/questions-accordion/QuestionsAccordion.styles'
+import Accordions from '../accordion/Accordions'
+
+interface QuestionsAccordion {
+  items: AccordionItem[]
+  title?: string
+}
+
+const QuestionsAccordion: FC<QuestionsAccordion> = ({ items, title = '' }) => {
```
MickhaelSh, line 18: bad name for component, it can be not only questions

MickhaelSh, line 18: also this component if it is reusable should provide props to acordions such as icon,sx,etc

mxrcury: maybe something like MultipleAccordionWithTItle?

src/components/accordion/Accordions.tsx
```diff
@@ -0,0 +1,98 @@
...
+  sx?: Sx
+  titleVariant: Variant
+  descriptionVariant: Variant
+}
+
+const Accordions: FC<AccordionsProps> = ({
+  items,
+  onChange,
+  activeIndex,
+  multiple = false,
```
MickhaelSh, line 35: we can omit this prop by adding variable multiple = Array.isArray(activeIndex)

src/components/questions-accordion/QuestionsAccordion.tsx
mxrcury: i think that this component is special only for offer details page and it to make this component reusable it will be needed one more wrapper

src/components/accordion/Accordions.tsx
```diff
@@ -0,0 +1,98 @@
...
+  icon,
+  sx = styles,
+  titleVariant,
+  descriptionVariant,
+  ...props
+}) => {
+  const { t } = useTranslation()
+
+  const isMultiple = multiple && activeIndex instanceof Array
```
MickhaelSh, line 44: this will be no needed

src/components/accordion/Accordions.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const accordionStyle = sx[icon ? 'withShowMoreIcon' : 'noShowMoreIcon']
+
+  return (
+    <Box sx={accordionStyle.root}>
+      {items.map((item, index) => (
+        <Accordion
+          data-testid={`${index}-${
+            isMultiple ? activeIndex.includes(index) : activeIndex === index
```
MickhaelSh, line 53: add this to variable in map, active = isMultiple ? activeIndex.includes(index) : activeIndex === index

src/components/accordion/Accordions.tsx
```diff
@@ -0,0 +1,98 @@
...
+import AccordionSummary from '@mui/material/AccordionSummary'
+import AccordionDetails from '@mui/material/AccordionDetails'
+
+import { styles } from '~/components/accordion/Accordion.styles'
+
+import { AccordionItem } from '~/types'
+import { Variant } from '@mui/material/styles/createTypography'
+
+interface Sx {
+  [key: string]: any
```
MickhaelSh, line 16: if we know keys which can be in this sx we should provide them instead of key:string

MickhaelSh (commented):
pay attention to code smells, and fix them if you have one

src/components/questions-accordion/QuestionsAccordion.tsx
MickhaelSh: yes something like this, use gpt it generate good names

MickhaelSh: if it is only for offer details it should be container not component, but I shink we will need such component in future, do it as component with multiple open accordions and use it for your needs on offer details

src/components/accordion/Accordions.tsx
mxrcury: don't you really think that it is unnecessary to type styles?

src/components/accordion/Accordions.tsx
mxrcury: i don't get what do you want, add field "active" to accordion items or what?

src/components/accordion/Accordions.tsx
mxrcury: refactored

src/components/accordion/Accordions.tsx
MickhaelSh: isMultiple ? activeIndex.includes(index) : activeIndex === index
 you use this code a few times, move it to variable called active

src/components/accordion/Accordions.tsx
MickhaelSh: no, it will be helpful for those who will use your component, they will know which keys in the object they can provide

src/components/accordion/Accordions.tsx
mxrcury: inside of map to make like  items.map((item, index) => { 
let active = false;
if(isMultiple) active = activeIndex.includes(index)
 }

MickhaelSh: nope you dont need let here, in first coment try to use ternary 
const active = isMultiple ? activeIndex.includes(index) : activeIndex === index
return (...JSX)

src/components/accordion/Accordions.tsx
mxrcury: it would be helpful if we have required styles for Accordions but there are no required fields besides noShowMoreIcon and withShowMoreIcon, nested fields with inside both of them are just by your choice and we can't know if in the future those who will be using this component will need to pass all styles that are in accordion

MickhaelSh: so noShowMoreIcon and withShowMoreIcon should be described in types  if they are required for now

src/components/accordion/Accordions.tsx
mxrcury: removed

src/components/accordion/Accordions.tsx
mxrcury: changed

src/components/accordion/Accordions.tsx
mxrcury: typed

src/components/multi-accordion-with-title/MultiAccordionWithTitle.styles.ts
```diff
@@ -0,0 +1,39 @@
...
+        },
+        titleInactive: {
+            ml: '13px',
+            color: 'primary.700'
+        },
+        description: {
+            ml: '27px'
+        }
+    },
+    noIcon: {}
```
MickhaelSh, line 38: find a way to omit using empty objects in styles

src/types/components/multi-accordion-with-title/multiAccordionWithTitle.interface.ts
```diff
@@ -0,0 +1,6 @@
+import { SxProps } from '@mui/material';
+import { AccordionSx } from '~/types';
+
+export interface MultiAccordionWithTitleSx extends AccordionSx {
+    [key: string]: SxProps
```
MickhaelSh, line 5: it has the same keys as accordionsx plus title

src/types/components/accordions/accordions.interface.ts
```diff
@@ -0,0 +1,13 @@
+import { SxProps } from "@mui/material"
+
+export interface AccordionItem {
+    title: string
+    description: string
+}
+interface NestedStyles {
```
MickhaelSh, line 7: don't use this , withIcon and noIcon have certain keys you can create interface for them

mxrcury: certain keys?for example?

src/components/multi-accordion-with-title/MultiAccordionWithTitle.styles.ts
mxrcury: how?it's no way if you want to see typed styles

src/types/components/accordions/accordions.interface.ts
mxrcury: certain interface is unnecessary, because all fields are partial

mxrcury: every field has no required for using in styles

MickhaelSh: what i see in your code is accordion,summary,titleActive,etc

mxrcury: all of them are optional for adding in styles so why need to write interface for this styles?

src/types/components/accordions/accordions.interface.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface AccordionItem {
+    title: string
+    description: string
+}
+interface NestedStyles {
+    [key: string]: string | number | NestedStyles
+}
+export interface AccordionSx {
+    withIcon: { [key: string]: SxProps | NestedStyles }
+    noIcon: { [key: string]: SxProps | NestedStyles }
```
MickhaelSh, line 12: do noIcon?: , so it will be not required, and you wont have to pass empty object

src/types/components/accordions/accordions.interface.ts
mxrcury: they will be needed to omit as you have written about noIcon

src/types/components/accordions/accordions.interface.ts
mxrcury: does it have a sense to type styles if field is not required and those who reuse it is choosing what to style just by checking component

src/types/components/accordions/accordions.interface.ts
MickhaelSh: because other fields which will can be provided to [key:string]:bla bla, will have no effect on your component, so it has to be typed

src/types/components/accordions/accordions.interface.ts
MickhaelSh: yes it have, styles as other props have to be typed

mxrcury: ok

src/components/accordion/Accordion.styles.ts
```diff
@@ -4,8 +4,16 @@ import {
   commonHoverShadow
 } from '~/styles/app-theme/custom-shadows'
 
+const noIconTitleStyles = {
+  fontSize: { md: '20px', sm: '13px' },
```
Tolik170, line 8: you can use h6 variant for md and i guess caption or body2 for sm

src/components/accordion/Accordion.styles.ts
```diff
@@ -21,11 +29,7 @@ export const styles = {
         display: 'none'
       }
     },
-    title: {
-      fontSize: { md: '20px', sm: '13px' },
-      lineHeight: { md: '28px', sm: '18px' },
-      color: 'primary.900'
-    },
+    titleActive: noIconTitleStyles, titleInactive: noIconTitleStyles,
```
Tolik170, line 32: "titleActive" and "titleInactive" have the same styles

src/components/accordion/Accordions.tsx
```diff
@@ -0,0 +1,93 @@
...
+import Typography from '@mui/material/Typography'
+import Box from '@mui/material/Box'
+import Accordion, { AccordionProps } from '@mui/material/Accordion'
+import AccordionSummary from '@mui/material/AccordionSummary'
+import AccordionDetails from '@mui/material/AccordionDetails'
+
+import { styles } from '~/components/accordion/Accordion.styles'
+
+import { AccordionItem, AccordionSx } from '~/types'
+import { Variant } from '@mui/material/styles/createTypography'
```
Tolik170, line 13: move it to the top

src/components/multi-accordion-with-title/MultiAccordionWIthTitle.tsx
```diff
@@ -0,0 +1,69 @@
...
+      multiple={true}
+      sx={{ withIcon: sx.withIcon, noIcon: sx.noIcon }}
+      titleVariant={VariantEnum.H6}
+      descriptionVariant={VariantEnum.Body2}
+      elevation={0}
+    />
+  )
+
+  return (
+    <Box sx={title ? sx.container : {}}>
```
Tolik170, line 62: try to title && sx.container

src/components/accordion/Accordions.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const { t } = useTranslation()
+
+  const isMultiple = Array.isArray(activeIndex)
+
+  const accordionStyle = sx[icon ? 'withIcon' : 'noIcon'] || {}
+
+  return (
+    <Box sx={accordionStyle.root}>
+      {items.map((item, index) => {
+        const active = isMultiple
```
Tolik170, line 46: move map above and pass a variable in here, for better readability

src/components/accordion/Accordions.tsx
MickhaelSh, line 13: in some files you use variantEnum from types. but here this from mui, you should stick to one approach

mxrcury: i used Variant as type from MUI just to type some field and VariantEnum as constants for adding values to typography, do you see it as a problem to use MUI types for typing input values in MUI component such as variant for Typography?

src/components/accordion/Accordions.tsx
mxrcury: i'm not sure that it will increase readability, cause this map is one what we have inside of this component

src/components/multi-accordion-with-title/MultiAccordionWIthTitle.tsx
mxrcury: it already was mentioned in another comments and i have answered to it as this way of doing it won't work, you can try by urself and assure, please check comments next time

src/components/accordion/Accordions.tsx
mxrcury: moved

src/components/accordion/Accordion.styles.ts
mxrcury: changed

src/components/accordion/Accordions.tsx
Tolik170: I think it will, return and map inside return doesn`t seem well

src/components/accordion/Accordions.tsx
MickhaelSh: no, it is not problem, but if we have enum for typography variants we can use it not only for adding value but also for type checking

src/components/accordion/Accordion.styles.ts
maarekua: These fields should be optional.

src/components/app-pagination/AppPagination.tsx
```diff
@@ -4,9 +4,11 @@ import Pagination, { PaginationProps } from '@mui/material/Pagination'
 import Box from '@mui/system/Box'
 
 import { styles } from '~/components/app-pagination/AppPagination.styles'
+import { SizeEnum } from '~/types'
 
 interface AppPaginationProps
   extends Omit<PaginationProps, 'count' | 'onChange'> {
+  size?: SizeEnum.Medium | SizeEnum.Large | SizeEnum.Small
```
maarekua, line 11: Shouldn't it be just SizeEnum type?

mxrcury: yeap, changed

PR #706, opened 2023-05-09
changed routes to kebab case

Tolik170:


Yurenko (approved):

src/constants/translations/en/header.json
```diff
@@ -1,13 +1,13 @@
 {
-  "whatCanYouDo": "What can you do",
-  "howItWorks": "How it works",
-  "whoWeAre": "Who we are",
+  "what-сan-you-do": "What can you do",
+  "how-it-works": "How it works",
+  "who-we-are": "Who we are",
```
Yurenko, line 4: I guess it doesn't need to establish kebab case here.
We need to change only routes.

Tolik170: in some files we mapped by routes and show for example `header.${item.route}, and  also, route should be same as path

mxrcury (approved):

maarekua (approved):

PR #713, opened 2023-05-09
created EnrollOffer component 

MickhaelSh:
added tests, translations for it, small fixes
image


src/components/app-text-area/AppTextArea.tsx
```diff
@@ -18,11 +18,19 @@ const AppTextArea: FC<AppTextAreaProps> = ({
   minRows = 4,
   maxRows = 4,
   maxLength,
+  title,
   value,
+  sx,
   ...props
 }) => {
+  const titleEl = title && (
+    <Typography sx={styles.title} variant='body2'>
```
mxrcury, line 27: i think it would be good to use enum of VariantEnum for variant property

Yurenko (approved):

src/components/app-text-area/AppTextArea.tsx
MickhaelSh: fixed

tsivkadi (approved):

maarekua (approved):

PR #716, opened 2023-05-10
Added block with comments to offer details page

Yurenko:
image
image


src/components/comment/Comment.styles.ts
```diff
@@ -41,18 +41,20 @@ export const styles = {
...
-    },
-    description: {
-      typography: 'overline',
-      color: 'primary.500'
+    titleWithDescription: {
+      title: {
+        display: 'flex',
+        flexDirection: { sm: 'column' },
+        columnGap: { xs: '12px' },
+        mb: 0,
```
MickhaelSh, line 49: those mb and m are not necessary now, i think

src/pages/offer-details/OfferDetails.tsx
```diff
@@ -60,6 +66,14 @@ const OfferDetails = () => {
           onBookmarkClick={onBookmarkClick}
         />
       </AppCard>
+      <AppCard sx={styles.wrapper}>
+        <ComentsBlock
```
MickhaelSh, line 70: you should add title as a props, because now we only have what students say, but we will use comment for students too

src/components/comment/Comment.styles.ts
Yurenko: Deleted

src/pages/offer-details/OfferDetails.tsx
Yurenko: Fixed

maarekua (approved):

PR #725, opened 2023-05-12
added about offer block to offerDetails page

Tolik170:
image
image


src/components/show-more-collapse/ShowMoreCollapse.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <Typography sx={styles.description(collapsedHeight)}>
+      {expanded ? description : description.slice(0, collapsedTextLength)}
+    </Typography>
+  )
+
+  return (
+    <Box>
+      <Typography sx={styles.title}>{title}</Typography>
+
+      <Collapse collapsedSize={collapsedSize} in={expanded} {...props}>
```
mxrcury, line 44: if you want to add here props that Collapse component from MUI receives, so it i think it needs to add extending of CollapseProps interface from MUI to ShowMoreCollapseProps

src/constants/translations/en/common.json
```diff
@@ -10,13 +10,16 @@
...
+  "showMore": "Show more",
+  "showLess": "Show less",
   "confirmButton": "OK",
   "uah": "UAH",
   "hour": "HOUR",
   "hourSlash": "/ HOUR",
   "beginner": "Beginner",
   "offers": "offers",
   "popularCategories": "Popular Categories",
+  "aboutOffer": "About Offer",
```
mxrcury, line 22: isn't it needed to add aboutOffer also to ua translations

src/components/show-more-collapse/ShowMoreCollapse.tsx
Tolik170: you are right, fixed

src/constants/translations/en/common.json
Tolik170: Didn`t see it, fixed

Yurenko (approved):

maarekua (approved):

PR #746, opened 2023-05-18
Fetch data on onFocus in the stepper

Yurenko:


MickhaelSh (approved):

src/containers/tutor-home-page/general-info-step/GeneralInfoStep.jsx
```diff
@@ -78,13 +78,19 @@ const GeneralInfoStep = ({
     service: getUserById,
     defaultResponse: { firstName: '', lastName: '' }
   })
-  const { response: countries } = useAxios({
+
+  const {
+    loading: loadingCountries,
+    response: countries,
+    fetchData: getCountriesFetch
```
Tolik170, line 85: i guess, you should not use both get and fetch in function name

Yurenko: You are right. I fixed it.

maarekua (approved):

PR #759, opened 2023-05-25
Fixed login button clickable on empty inputs

TSlashDreamy:
The login button can be clicked without entering any data in the Email and Password field has been fixed by adding a check for empty fields and data validity.

Tests report:
зображення

### DIFF
--- a/src/containers/guest-home-page/login-form/LoginForm.jsx
+++ b/src/containers/guest-home-page/login-form/LoginForm.jsx
@@ -1,4 +1,3 @@
-import { useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import useInputVisibility from '~/hooks/use-input-visibility'
 import { useSelector } from 'react-redux'
@@ -32,8 +31,6 @@ const LoginForm = ({
   const openForgotPassword = () => {
     openModal({ component: <ForgotPassword /> })
   }
-
-  const inputsAreValid = useMemo(() => data.email && data.password)
 
   return (
     <Box component='form' onSubmit={handleSubmit}>
@@ -73,7 +70,7 @@ const LoginForm = ({
       </Typography>
 
       <AppButton
-        disabled={!inputsAreValid}
+        disabled={!data.email || !data.password}
         loading={authLoading}
         sx={styles.loginButton}
         type='submit'
--- a/src/containers/guest-home-page/login-form/LoginForm.jsx
+++ b/src/containers/guest-home-page/login-form/LoginForm.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react'
+import { useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import useInputVisibility from '~/hooks/use-input-visibility'
 import { useSelector } from 'react-redux'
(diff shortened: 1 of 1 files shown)


src/containers/guest-home-page/login-form/LoginForm.jsx
```diff
@@ -31,6 +35,17 @@ const LoginForm = ({
     openModal({ component: <ForgotPassword /> })
   }
 
+  useEffect(() => {
```
MickhaelSh, line 38: use useMemo instead useEffect and useState

src/containers/guest-home-page/login-form/LoginForm.jsx
```diff
@@ -69,6 +84,7 @@ const LoginForm = ({
       </Typography>
 
       <AppButton
+        disabled={!inputsIsValid}
         loading={authLoading}
         size='large'
```
MickhaelSh, line 89: its large by default, remove

MickhaelSh: and variant contained too

src/containers/guest-home-page/login-form/LoginForm.jsx
```diff
@@ -31,6 +35,17 @@ const LoginForm = ({
     openModal({ component: <ForgotPassword /> })
   }
 
+  useEffect(() => {
+    const { email: emailError, password: passwordError } = errors
+    const { email: emailData, password: passwordData } = data
+
+    if (!emailError && !passwordError && emailData && passwordData) {
```
tsivkadi, line 42: as i remember, button should be disabled only when fields are empty

src/containers/guest-home-page/login-form/LoginForm.jsx
TSlashDreamy: This is really a better solution, thanks for your advice, implemented it

src/containers/guest-home-page/login-form/LoginForm.jsx
TSlashDreamy: Removed unnecessary lines 👍

src/containers/guest-home-page/login-form/LoginForm.jsx
TSlashDreamy: Yeah, you are right 😅. Fixed it

Yurenko (approved):

src/containers/guest-home-page/login-form/LoginForm.jsx
```diff
@@ -31,6 +33,8 @@ const LoginForm = ({
     openModal({ component: <ForgotPassword /> })
   }
 
+  const inputsAreValid = useMemo(() => data.email && data.password)
```
MickhaelSh, line 36: useMemo without dependency does nothing, and as I see we don`t need even it in this code. so you can just put this code in button, disabled={!data.email || !data.password}

TSlashDreamy: You are right, it will be much better that way. Sorry for the silly mistakes, I still have a lot to learn 😅. Fixed it

tsivkadi (approved):

maarekua (approved):

PR #762, opened 2023-05-25
Fixed password confirmation bug

debikus:
Password confirmation field now will check matching to password field

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ test-report.xml
 /.idea
 .idea
 .DS_Store
+.env
 .env.local
 .env.development.local
 .env.test.local
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,6 @@ test-report.xml
 /.idea
 .idea
 .DS_Store
-.env
 .env.local
 .env.development.local
 .env.test.local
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -19,6 +19,7 @@ test-report.xml
 /.idea
 .idea
 .DS_Store
+.env
```
MickhaelSh, line 22: remove this, use .env.local

src/utils/validations/login.js
```diff
@@ -19,6 +19,7 @@ export const lastName = (value) => {
 export const confirmPassword = (password, data) => {
   return emptyField(
     password,
+    !password ? 'common.errorMessages.emptyField' : '',
```
MickhaelSh, line 22: why do you need this check here if empty field func has check for value ?

src/constants/translations/en/common.json
```diff
@@ -69,4 +69,4 @@
     "spanish": "Spanish",
     "arabic": "Arabic"
   }
-}
+}
```
MickhaelSh, line 72: why have you pushed this file, if you changed nothing in it

Tolik170: lint added required space for this file, so, I see no reason to revert changes in this file now

maarekua (approved):

PR #766, opened 2023-05-26
Fixed space error between two words

debikus:
Now user can use spacebar between words
image

### DIFF
--- a/src/utils/validations/common.ts
+++ b/src/utils/validations/common.ts
@@ -10,7 +10,7 @@ const validations: Validations = {
     if (value.length > 30) {
       return 'common.errorMessages.nameLength'
     }
-    if (!RegExp(/^[a-zа-яєії]+$/i).test(value)) {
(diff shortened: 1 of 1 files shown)


src/utils/validations/common.ts
```diff
@@ -66,14 +66,14 @@ export const numberField = (value: string, errorMessage: string) => {
 
 export const textField =
   (min: number, max: number) =>
-  (value: string): string | undefined => {
-    if (value.length !== 0 && value.length < min) {
-      return 'common.errorMessages.shortText'
+    (value: string): string | undefined => {
```
debikus, line 69: that changed because vscode automatically tabulate after saving file

Tolik170 (approved):

qavs:
Re-tested, bug is still there. Both fields (First name, Last name) don't validate spase between two separate names.
image

PR #779, opened 2023-05-27
The “What can you do” button is not highlighted after clicking it, fixed.

TSlashDreamy:
I added a `:focus` pseudo-class to solve this problem. NavItem is highlighted on click now but if we click another item on the page then the highlight is removed from the item. I am not sure is this supposed element behavior.

Result:

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/45912519/c43c37fd-f103-4550-828b-7bd55cef0e5c

### DIFF
--- a/src/containers/layout/navbar/NavBar.styles.js
+++ b/src/containers/layout/navbar/NavBar.styles.js
@@ -17,8 +17,11 @@ export const styles = {
   navItemText: {
     color: 'primary.900',
     textDecoration: 'none',
-    '&:focus': {
+    '&:hover': {
       color: 'primary.500'
+    },
+    '&:focus': {
+      textDecoration: 'underline'
     }
   },
   navItem: {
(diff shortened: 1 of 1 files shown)


Yurenko (approved):

tsivkadi (approved):

src/containers/layout/navbar/NavBar.styles.js
```diff
@@ -16,7 +16,10 @@ export const styles = {
   },
   navItemText: {
     color: 'primary.900',
-    textDecoration: 'none'
+    textDecoration: 'none',
+    '&:focus': {
```
MickhaelSh, line 20: I`m not sure is focus well in this case, because after scrolling out of the section, the button continues to be highlighted, we need to discuss this in meeting, I think hover would be better here

maarekua (approved):

src/containers/layout/navbar/NavBar.styles.js
TSlashDreamy: > I`m not sure is focus well in this case, because after scrolling out of the section, the button continues to be highlighted, we need to discuss this in meeting, I think hover would be better here

We discussed it. Made some changes 👍

Issue #688, opened 2023-05-05
[Tutor "Sign Up" pop-up] the error message is shown, when the 'First name' field data contains space between two words


qavs:
**Description**
The error message is shown, when the 'First name' field data contains space between two words.

**Environment:** Windows 10 21H2, Chrome 110.0.5481.178
**Defect Type:** Functional
**Priority:** Medium
**Severity**: Minor
**Reproducible:** Always

**Preconditions:**
The https://s2s-front-stage.azurewebsites.net/ site is opened.
The button "Become a Tutor" is clicked and "Sign Up" pop-up appeared.

**Steps to reproduce:**

Fill the text field ‘First name’ with valid data.(e.g. 'Valeriia Maria')

**Actual result:**

The error 'Field can contain alphabetic characters only' is shown.
![image](https://user-images.githubusercontent.com/129952218/236569416-f3bdbc3f-7398-42f3-a35c-7ecb6c4ac04c.png)

**Expected result:**
Transparent header disappears, system accepts two words with a space between them as valid option, due to requirements.

PR #780, opened 2023-05-28
fixed bug with input

VladyslavPetskovych:
[Tutor "Sign Up" pop-up] the error message is shown, when the 'First name' field data contains space between two words #688

Simply added the space to regex exeptions.
image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -26,6 +26,7 @@ module.exports = {
     }
   },
   rules: {
+    'prettier/prettier': 0,
     'prefer-arrow-callback': ['error'],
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
     }
   },
   rules: {
+    'prettier/prettier': 0,
```
MickhaelSh, line 29: remove

.env
MickhaelSh, line 1: remove this file

MickhaelSh (commented):
in the description result of the fix, not tests or pre-commit

PR #783, opened 2023-05-29
Changed some files to ts

MickhaelSh:
Changed some files to ts
image


Tolik170 (approved):

src/utils/image-resize.ts
```diff
@@ -1,4 +1,7 @@
-export const imageResize = async (photoPath, { newHeight, newWidth }) => {
+export const imageResize = async (
+  photoPath: string,
+  { newHeight, newWidth }: { newHeight: number; newWidth: number }
```
mxrcury, line 3: maybe create separate interface for this object type?

PR #785, opened 2023-05-29
disable closing window outside popup

DLmar:
We needed to prevent window closing when we click outside our modal window. Now it's fixed

### DIFF
--- a/src/components/popup-dialog/PopupDialog.tsx
+++ b/src/components/popup-dialog/PopupDialog.tsx
@@ -45,6 +45,7 @@ const PopupDialog: FC<PopupDialogProps> = ({
       data-testid='popup'
       fullScreen={isMobile}
       maxWidth='xl'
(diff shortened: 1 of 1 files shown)


src/components/popup-dialog/PopupDialog.tsx
```diff
@@ -45,6 +45,7 @@ const PopupDialog: FC<PopupDialogProps> = ({
       data-testid='popup'
       fullScreen={isMobile}
       maxWidth='xl'
+      onClick={(e) => e.stopPropagation()}
```
Tolik170, line 48: I guess, we should add leave confirmation when data is filled in the modal, not disable default behaviour

Tolik170 (changes requested):

Tolik170 (approved):

maarekua (approved):

PR #806, opened 2023-05-31
The “Space2Study” logo button does not return to the home page welcoming block after clicking it fixed

TSlashDreamy:
Only the GuestHomepage component is scrollable, so the standard method `window.scrollTo(0, 0)` does not work for it. To solve this problem, I used the hash navigation as in navlinks.

Result:

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/45912519/3f5768f0-e0b0-4650-b8ce-1119c5c59917

### DIFF
--- a/src/containers/layout/navbar/NavBar.jsx
+++ b/src/containers/layout/navbar/NavBar.jsx
@@ -1,6 +1,5 @@
 import { useState, useEffect } from 'react'
 import { useSelector } from 'react-redux'
-import { Link } from 'react-router-dom'
 import { guestRoutes } from '~/router/constants/guestRoutes'
 import { studentRoutes } from '~/router/constants/studentRoutes'
 import { tutorRoutes } from '~/router/constants/tutorRoutes'
@@ -60,10 +59,10 @@ const Navbar = () => {
   return (
     <Box sx={styles.header}>
       <Button
-        component={Link}
+        component={HashLink}
         size='small'
         sx={styles.logoButton}
-        to={guestRoutes.home.path}
+        to={guestRoutes.home.path + '#welcome'}
       >
         <Logo />
       </Button>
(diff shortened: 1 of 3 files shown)


src/containers/layout/navbar/NavBar.jsx
```diff
@@ -60,10 +59,10 @@ const Navbar = () => {
   return (
     <Box sx={styles.header}>
       <Button
-        component={Link}
+        component={HashLink}
         size='small'
         sx={styles.logoButton}
-        to={guestRoutes.home.path}
+        to={guestRoutes.home.path + '#welcome'}
```
MickhaelSh, line 65: hash should be added to routes, not hardcoded in component

TSlashDreamy: added it to route constants instead

maarekua (approved):

src/router/constants/guestRoutes.ts
```diff
@@ -1,5 +1,5 @@
 export const guestRoutes = {
-  home: { route: '/', path: '/' },
+  home: { route: '/', path: '/#welcome' },
```
MickhaelSh, line 2: not here, home path should be clear

TSlashDreamy: > not here, home path should be clear

cleared home path and created separate route for this

src/containers/layout/navbar/NavBar.jsx
```diff
@@ -60,10 +59,10 @@ const Navbar = () => {
   return (
     <Box sx={styles.header}>
       <Button
-        component={Link}
+        component={HashLink}
         size='small'
         sx={styles.logoButton}
-        to={guestRoutes.home.path}
+        to={guestRoutes.home.path + guestRoutes.welcome.path}
```
MickhaelSh, line 65: u can do like routs are done in navbar so concatination will be no needed

TSlashDreamy: > u can do like routs are done in navbar so concatination will be no needed

left only one route path as done in the navbar

PR #807, opened 2023-05-31
[Home page] The slash separator between home page buttons is not corresponded to the mockup

DLmar:
From:
image
To:
image


src/styles/index.css
```diff
@@ -1,4 +1,5 @@
 @import url('https://fonts.googleapis.com/css2?family=Rubik:wght@300;400;500;600;800;900&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Poppins:ital,wght@0,100;0,200;0,300;0,400;0,500;0,600;0,700;0,800;0,900;1,100;1,200;1,300;1,400;1,500;1,600;1,700;1,800;1,900&display=swap');
```
MickhaelSh, line 2: we use Rubick accross app, remove this

MickhaelSh: image
Yes, no need to put another font for slashes

DLmar: Ok, now it will look like this. Is this ok ?
image

maarekua (approved):

PR #810, opened 2023-06-01
(SP: 1) Write unit test for tab

DLmar:


tests/unit/components/tab/Tab.spec.jsx
```diff
@@ -0,0 +1,37 @@
+import { render, screen, fireEvent } from '@testing-library/react'
+import Tab from '~/components/tab/Tab'
+
+describe('Tab', () => {
+  it('renders the label', () => {
+    const label = 'label text'
+    render(<Tab activeTab label={label} onClick={() => {}} />)
+    const tabLabel = screen.getByText(label)
```
FryDay969, line 8: I recommend you move out the "label" variable and the onClick function from the test. Besides that, you may move out from the test "render" method and put in the beforeEach() constructions. These actions will help you to avoid redundant code.

tests/unit/components/tab/Tab.spec.jsx
```diff
@@ -0,0 +1,37 @@
...
+  it('calls onClick function when clicked', () => {
+    let onClickCalled = false
+    const onClick = () => {
+      onClickCalled = true
+    }
+    const label = 'label text'
+    render(<Tab activeTab label={label} onClick={onClick} />)
+    const tabButton = screen.getByRole('button')
+    fireEvent.click(tabButton)
+    expect(onClickCalled).toBe(true)
```
tsivkadi, line 35: use toHaveBeenCalled method here

tests/unit/components/tab/Tab.spec.jsx
```diff
@@ -0,0 +1,37 @@
+import { render, screen, fireEvent } from '@testing-library/react'
+import Tab from '~/components/tab/Tab'
+
+describe('Tab', () => {
+  it('renders the label', () => {
+    const label = 'label text'
+    render(<Tab activeTab label={label} onClick={() => {}} />)
```
tsivkadi, line 7: mock onClick fn

tests/unit/components/tab/Tab.spec.jsx
```diff
@@ -0,0 +1,37 @@
...
+    const tabLabel = screen.getByText(label)
+    expect(tabLabel).toBeInTheDocument()
+  })
+
+  it('applies activeTab if it is true', () => {
+    const label = 'label text'
+    render(<Tab activeTab label={label} onClick={() => {}} />)
+    const tabButton = screen.getByRole('button')
```
mxrcury, line 15: i think it would be clearer to move this text 'button' in separate value and call something like 'tabButtonRole'

tests/unit/components/tab/Tab.spec.jsx
```diff
@@ -0,0 +1,37 @@
...
+  it('not applies activeTab if it is false', () => {
+    const label = 'label text'
+    render(<Tab activeTab={false} label={label} onClick={() => {}} />)
+    const tabButton = screen.getByRole('button')
+    expect(tabButton).not.toHaveClass('activeTab')
+  })
+
+  it('calls onClick function when clicked', () => {
+    let onClickCalled = false
+    const onClick = () => {
```
mxrcury, line 28: for mocked functions it is recommended to use mocking function from vitest, i mean `vi.fn()`

maarekua (approved):

tests/unit/components/tab/Tab.spec.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+  it('applies activeTab if it is true', () => {
+    const tabButton = screen.getByRole('button')
+    if (tabButton.classList.contains('activeTab')) {
+      expect(tabButton).toHaveClass('activeTab')
+    }
+  })
+
+  it('does not apply activeTab if it is false', () => {
+    const tabButton = screen.getByRole('button')
```
Tolik170, line 29: add spaces between fireEvents and expects

tests/unit/components/tab/Tab.spec.jsx
```diff
@@ -0,0 +1,38 @@
...
+import { vi } from 'vitest'
+
+import Tab from '~/components/tab/Tab'
+
+describe('Tab', () => {
+  let label
+  let onClick
+
+  beforeEach(() => {
+    label = 'label text'
```
tsivkadi, line 11: put variables outside the test, there is no need for it to be in beforeEach

FryDay969 (approved):

tests/unit/components/tab/Tab.spec.jsx
DLmar: i decided to leave as it is

PR #838, opened 2023-06-06
fixed the text under the button Sign Up with Google

Taras-Shavel:
I changed the text from "Already have a tutor account? Login!" to "Already have a Space2Study account?" under the button Sign Up with Google
Inkedimage


.env.local
Yurenko, line 1: Don't push env file.

Yurenko (changes requested):

tsivkadi (approved):

.env.local
Taras-Shavel: done

Yurenko (approved):

PR #851, opened 2023-06-10
Change maxLength of professional status

DLmar:


Tolik170 (approved):

src/containers/tutor-home-page/general-info-step/GeneralInfoStep.jsx
```diff
@@ -205,7 +205,7 @@ const GeneralInfoStep = ({
           <AppTextArea
             fullWidth
             label={t('becomeTutor.generalInfo.textFieldLabel')}
-            maxLength={70}
+            maxLength={200}
```
Tolik170, line 208: Change length in validations too: 
image

Yurenko: You can see an error message when you type more than 70 letters.
image

Tolik170: Maybe we should remove this validation as it will not work out for the same length as on import

MickhaelSh (approved):

tsivkadi (approved):

Issue #616, opened 2023-04-21
[Home page] The buttons are not centered at the home page footer.


AntonOkun:
**Description:** The buttons are not centered at the home page footer according to the mockup.

**Environment:** Windows 10 x64, Microsoft Edge 112.0.1722.48, Google Chrome 112.0.5615.122

**Reproducible:** always.

**Build found:**

**Preconditions:**
The user is not logged in.

**Steps to reproduce:**
1. Open the https://s2s-front-stage.azurewebsites.net/ site.
2. Click on the “Who we are” button on the home page header.
3. Scroll down to the home page footer.
4. Take a look at the home page footer.

**Actual result:**
The buttons are not centered at the home page footer according to the mockup.
![RigPxbmBme](https://user-images.githubusercontent.com/129941062/233657612-bdadd2c5-a081-49ce-b593-087fc3c29c12.png)

**Expected result:**
The buttons are center aligned at the home page footer according to the mockup.
![hlc1w8vAHJ](https://user-images.githubusercontent.com/129941062/233694996-b0570ddb-a097-4e4a-bf1f-81dde9072a83.png)

**User story and test case links**
User story Home page footer #892 
[Test case](https://jira.softserve.academy/browse/LVISTQB-326)

PR #855, opened 2023-06-12
changed position for links in footer

VladyslavPetskovych:
changed position for links in footer
[Home page] The buttons are not centered at the home page footer. #616

### DIFF
--- a/src/containers/layout/footer/Footer.styles.ts
+++ b/src/containers/layout/footer/Footer.styles.ts
@@ -1,19 +1,24 @@
 export const styles = {
   footer: {
     backgroundColor: 'primary.900',
-    color: 'primary.50',
-    marginTop: 'auto'
+    color: 'primary.50'
   },
   container: {
     display: 'flex',
+    position: 'relative',
     flexDirection: { xs: 'column-reverse', sm: 'row' },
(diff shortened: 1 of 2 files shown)


src/containers/layout/footer/Footer.styles.ts
```diff
@@ -1,19 +1,24 @@
...
   container: {
     display: 'flex',
+    position: 'relative',
     flexDirection: { xs: 'column-reverse', sm: 'row' },
     justifyContent: { xs: 'space-between', sm: 'space-around' },
     alignItems: 'center',
     py: { xs: '12px', sm: '26px' }
   },
+  title: {
+    position: { xs: 'static', sm: 'absolute' },
```
Tolik170, line 15: It is not a great idea to use absolute with negative values, try to center it with flex or margin, also, notice that you have space around in container

src/containers/layout/footer/guest-footer/GuestFooter.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
MickhaelSh, line 1: we dont disable eslint, we fix warnings and errors

PR #874, opened 2023-06-15
Changed max width of guest page

MickhaelSh:
Changed max width of guest page, refactored and moved component to ts
image
image


src/containers/guest-home-page/WhatCanYouDo.tsx
```diff
@@ -3,62 +3,55 @@ import { useTranslation } from 'react-i18next'
...
+    actionType: UserRoleEnum.Tutor
   }
 ]
 
 const WhatCanYouDo = () => {
   const { t } = useTranslation()
   const { openModal } = useModalContext()
 
-  const openDialog = (type) => {
+  const openDialog = (type: string) => {
```
mxrcury, line 37: as i have seen type value can be only one of user roles so maybe type it with role enum or united type of roles, how do you think?

Yurenko (approved):

tsivkadi (approved):

src/containers/guest-home-page/WhatCanYouDo.tsx
MickhaelSh: indeed. fixed

Tolik170 (approved):

maarekua (approved):

PR #948, opened 2023-07-17
Added spaces between description blocks

TSlashDreamy:
Preview
зображення

<details>
<summary>Lint</summary>

зображення

</details>

### DIFF
--- a/src/components/carousel-with-image/CarouselWithImage.styles.ts
+++ b/src/components/carousel-with-image/CarouselWithImage.styles.ts
@@ -17,7 +17,8 @@ export const styles = {
     minHeight: '144px',
     backgroundColor: 'primary.800',
     borderRadius: '6px',
-    p: '16px'
+    p: '16px',
+    m: '0 6px'
   },
   steper: {
     display: 'flex',
--- a/src/components/carousel-with-image/CarouselWithImage.styles.ts
+++ b/src/components/carousel-with-image/CarouselWithImage.styles.ts
@@ -18,7 +18,7 @@ export const styles = {
     backgroundColor: 'primary.800',
     borderRadius: '6px',
     p: '16px',
-    m: '0 6px'
+    m: '0 1.5vw'
   },
   steper: {
     display: 'flex',


FryDay969 (dismissed):

src/components/carousel-with-image/CarouselWithImage.styles.ts
```diff
@@ -17,7 +17,8 @@ export const styles = {
     minHeight: '144px',
     backgroundColor: 'primary.800',
     borderRadius: '6px',
-    p: '16px'
+    p: '16px',
+    m: '0 6px'
```
Yurenko, line 21: Here we use display: flex, so you can add gap property instead of margin. 
I think it should work.

TSlashDreamy:
> image not enough space for 540px for example

thanks, you are right, this is because svg itself have the whitespaces that cannot be edited and the image are resizing responsively,  when the window resizes. Fixed it by using vw instead of px

src/components/carousel-with-image/CarouselWithImage.styles.ts
TSlashDreamy: > Here we use display: flex, so you can add gap property instead of margin. I think it should work.

It's actually a great solution to add gap inside a box, but not to add gaps outside it. Svg in the component have "fake" whitespaces that cannot be edited, so basically, i think, adding margins to the description box it's not so bad solution

DLmar (approved):

FryDay969 (approved):

src/components/carousel-with-image/CarouselWithImage.styles.ts
```diff
@@ -17,7 +17,8 @@ export const styles = {
     minHeight: '144px',
     backgroundColor: 'primary.800',
     borderRadius: '6px',
-    p: '16px'
+    p: '16px',
+    m: '0 1.5vw'
```
Tolik170, line 21: I guess, it should have a static value, which does not depend on window size

TSlashDreamy: > I guess, it should have a static value, which does not depend on window size

As i understand, we have responsive svg above description block with "fake" white spaces (white spaces inside svg). 

If we use static value we got next result where gaps between blocks is not the same as between svgs. 
static

But if we use vw, this resolve the problem
adaptive

Tolik170 (approved):

Yurenko (approved):

maarekua (approved):

PR #1063, opened 2023-08-25
implemented rename functionality for attachment

Tolik170:
https://github.com/ita-social-projects/SpaceToStudy-Client/assets/63456632/6053dbdb-15a4-47b7-8c35-28eb403cec1a


src/types/attachment/interfaces/attachment.interface.ts
```diff
@@ -7,3 +7,8 @@ export interface Attachment extends CommonEntityFields {
   description: string
   size: number
 }
+
+export interface UpdateAttachmentParams {
+  fileName: Attachment['fileName']
+  id: Attachment['author']['_id']
```
Yurenko, line 13: use Pick<Attachment, 'fileName'> instead of 'Attachment['fileName']'

src/components/icon-extention-with-title/IconExtentionWithTitle.tsx
```diff
@@ -1,38 +1,93 @@
...
+  }
+
+  const onCancelHandler = () => {
+    onCancel?.()
+  }
+
+  const onSaveHandler = () => {
+    const trimmedValue = inputValue.trim()
+    const value = trimmedValue !== initialInputValue ? trimmedValue : ''
+    onSave && void onSave(value)
```
Yurenko, line 52: I guess, that you can leave just 
`const value = trimmedValue ? inputValue.trim() : ''`

Tolik170: I check if trimmedValue is equal to initial, so, if user didn`t change the name, I set it to an empty string, and if value is an empty string the new request will not be called

src/constants/translations/en/my-resources-page.json
```diff
@@ -22,7 +22,8 @@
     "file": "File",
     "size": "Size",
     "lastUpdate": "Last update",
-    "emptyAttachments": "You have no attachments yet"
+    "emptyAttachments": "You have no attachments yet",
+    "validationError": "File name cannot be longer than 55 symbol."
```
Yurenko, line 26: I guess you can use `File name cannot be longer than {{number}} symbol.`.
It is necessary to pass amount of symbols. As for me, it is reusable and we change number in one place.

src/components/icon-extention-with-title/IconExtentionWithTitle.tsx
Yurenko: Perhaps it is unnecessary to permit saving a file without a name.

Tolik170: It will not save the file without a name or without changing, onSave function check inside if the value is empty, if not, it will make a request

Roman-Peretiatko (approved):

TSlashDreamy (approved):

ArturBekhDEV (approved):

maarekua (approved):

PR #1109, opened 2023-09-04
Added redirect from dialog to chat

TSlashDreamy:
<details>
<summary>Preview</summary>

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/45912519/c93d6780-1d0c-46f6-b994-de87d5ca4bc2

</details>
<details>
<summary>Lint</summary>

зображення
</details>

## Update (08.09.23) 442a42b78e6261eeb08729c705a4bd8d55e583e4
<details>
<summary>Preview</summary>

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/45912519/827f0152-a187-4c89-9994-e7f4b78bf060

</details>
<details>
<summary>Lint</summary>

зображення

</details>


src/pages/chat/Chat.tsx
```diff
@@ -92,6 +94,14 @@ const Chat = () => {
     }
   }, [selectedChat, messages.length])
 
+  useEffect(() => {
+    if (currentChatId) {
+      listOfChats.forEach((chat: ChatResponse) => {
+        if (chat._id === currentChatId) setSelectedChat(chat)
```
Tolik170, line 100: As I see it will navigate only if chat exists?

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
```diff
@@ -6,6 +6,7 @@ import {
   useRef,
   FC
 } from 'react'
+import { useNavigate } from 'react-router-dom'
```
Tolik170, line 9: same here, move it above

src/pages/chat/Chat.tsx
TSlashDreamy: > As I see it will navigate only if chat exists?

Yup, that's right

src/containers/chat/chat-item/ChatItem.tsx
```diff
@@ -10,6 +10,7 @@ import { useAppSelector } from '~/hooks/use-redux'
 import { styles } from '~/containers/chat/chat-item/ChatItem.styles'
 import { ChatResponse, ComponentEnum, OverlapEnum, PositionEnum } from '~/types'
 import { getFormattedDate } from '~/utils/helper-functions'
+import { useChatContext } from '~/context/chat-context'
```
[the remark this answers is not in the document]
TSlashDreamy: > better set this import above near hooks

thanks for your advice, moved above near custom hooks

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
TSlashDreamy: > same here, move it above

thanks for your advice, moved above near custom hooks

src/pages/chat/Chat.tsx
Tolik170: It also should create a new chat and navigate to it if the chat doesn`t exist

TSlashDreamy: > It also should create a new chat and navigate to it if the chat doesn`t exist

I guess it wouldn't make sense, because we don't even have a button, unless we're talking to the user. 
зображення


And the chat will be created as soon as we send the first message
зображення

src/pages/chat/Chat.tsx
```diff
@@ -92,6 +94,14 @@ const Chat = () => {
     }
   }, [selectedChat, messages.length])
 
+  useEffect(() => {
+    if (currentChatId) {
+      listOfChats.forEach((chat: ChatResponse) => {
+        if (chat._id === currentChatId) return setSelectedChat(chat)
+      })
```
Yurenko, line 101: Here you use 'return' it is useless. 
And use maybe for loop instead of forEach, because when you find a certain chat and set chat in setSelectedChat(chat) you can break iteration.
forEach doesn't work break.

For example: It will be the first chat but you have fifty chats and when you use forEach it will run through the entire array.

TSlashDreamy: > Here you use 'return' it is useless. And use maybe for loop instead of forEach, because when you find a certain chat and set chat in setSelectedChat(chat) you can break iteration. forEach doesn't work break.
> 
> For example: It will be the first chat but you have fifty chats and when you use forEach it will run through the entire array.

Oh, you are absolutely right. Thank you for your advice! I think, using `listOfChats.find()` would be better than `listOfChats.for()`, because, `listOfChats.find()` performance wise is `O(n)`, and it would look much better in my opinion.

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
```diff
@@ -89,14 +101,40 @@ const ChatDialogWindow: FC<ChatDialogWindow> = ({ chatInfo }) => {
...
+      <Typography key={el.question} sx={styles.question}>
+        {t(el.question)}
+      </Typography>
+    ))
+  )
 
   const closeChatWindow = () => setChatInfo(null)
 
+  const openChatInNewTab = (chatId: unknown) => {
+    localStorage.setItem('currentChatId', chatId as string)
```
Yurenko, line 120: I guess you don't need to set id in localStorage. You can try to take it from context.

src/pages/chat/Chat.tsx
```diff
@@ -145,7 +160,7 @@ const Chat = () => {
         files={mockFiles}
         links={mockLinks}
         media={mockMedia}
-        member={selectedChat.members[0]}
+        member={selectedChat.members[1]}
```
mxrcury, line 163: should we search for user by specific data but not by index in the case when user that you need is under ***1*** index

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
```diff
@@ -89,14 +101,40 @@ const ChatDialogWindow: FC<ChatDialogWindow> = ({ chatInfo }) => {
...
+
+  const handleCreateNewChat = async () => {
+    setCreatingChat(true)
+    const newChat = await createChat()
+    openChatInNewTab(newChat.data._id)
+    setCreatingChat(false)
+  }
+
+  const handleRedirectToChat = async () => {
+    const chatId: unknown = chatInfo.chatId
```
mxrcury, line 132: i think it will be better to don't use `unknown` type and there is no need to type constant variable that get value from typed object

TSlashDreamy: > i think it will be better to don't use `unknown` type and there is no need to type constant variable that get value from typed object

yeah, that would be better. Fixed it 👍

src/pages/chat/Chat.tsx
TSlashDreamy: > should we search for user by specific data but not by index in the case when user that you need is under _**1**_ index

Oh, you're right, there were problems with that. Fixed it
Before:
зображення

After:
зображення

src/containers/chat/chat-item/ChatItem.tsx
```diff
@@ -26,29 +32,41 @@ const ChatItem: FC<ItemOfChatProps> = ({
...
   const { t } = useTranslation()
   const { userId } = useAppSelector((state) => state.appMain)
+  const userToSpeak = useMemo<Member | undefined>(
+    () => chat?.members.find((member) => member.user._id !== userId),
+    [chat, userId]
+  )
 
-  const { firstName, lastName, photo } = chat.members[0].user
-  const { text, author, updatedAt } = chat.latestMessage
+  const firstName = userToSpeak?.user?.firstName
```
Tolik170, line 40: do you really need a "?" character after user, I think userToSpeak?.user.firstName will be enough

also, you can destructure user from userToSpeak instead of creating these 3 variables

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
```diff
@@ -89,14 +101,40 @@ const ChatDialogWindow: FC<ChatDialogWindow> = ({ chatInfo }) => {
...
 
+  const openChatInNewTab = (chatId: unknown) => {
+    localStorage.setItem('currentChatId', chatId as string)
+    window.open(authRoutes.chat.path, '_blank', 'noopener noreferrer')
+  }
+
+  const handleCreateNewChat = async () => {
+    setCreatingChat(true)
+    const newChat = await createChat()
+    openChatInNewTab(newChat.data._id)
```
Tolik170, line 127: what if there will be an error during the request? Maybe make sense to use useAxios hook for handling an error and take loading from it, check it

src/containers/chat/chat-item/ChatItem.tsx
TSlashDreamy: you are right, removed it

src/containers/chat/chat-item/ChatItem.tsx
```diff
@@ -26,29 +32,41 @@ const ChatItem: FC<ItemOfChatProps> = ({
...
+      options: {
+        day: '2-digit',
+        month: '2-digit',
+        year: 'numeric'
+      },
+      isCurrentDayHours: true
+    })
 
-  const isCurrentUser = userId === author._id && (
+  const isCurrentUser = author && userId === author._id && (
```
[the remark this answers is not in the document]
TSlashDreamy: > do you need this check for author? because as I remeber this variable is true when my message is latest

nope. we don't. Removed

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
```diff
@@ -89,14 +101,40 @@ const ChatDialogWindow: FC<ChatDialogWindow> = ({ chatInfo }) => {
...
+    questions.map((el) => (
+      <Typography key={el.question} sx={styles.question}>
+        {t(el.question)}
+      </Typography>
+    ))
+  )
 
   const closeChatWindow = () => setChatInfo(null)
 
+  const openChatInNewTab = (chatId: unknown) => {
```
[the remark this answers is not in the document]
TSlashDreamy: > it should be number or string but not unknown

changed

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
TSlashDreamy: > what if there will be an error during the request? Maybe make sense to use useAxios hook for handling an error and take loading from it, check it

You are right. Fixed!

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/45912519/20dc456e-3ad6-48a9-a2e6-41f767df6ea9

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
```diff
@@ -68,6 +97,33 @@ const ChatDialogWindow: FC<ChatDialogWindow> = ({ chatInfo }) => {
...
   }, [chatInfo.chatId, messagesLoad])
 
+  useEffect(() => {
+    if (createdChat._id) openChatInNewTab(createdChat._id as string)
+  }, [createdChat])
+
+  const closeChatWindow = () => setChatInfo(null)
+
+  const openChatInNewTab = (chatId: Pick<ChatResponse, '_id'> | string) => {
+    localStorage.setItem('currentChatId', chatId as string)
```
Tolik170, line 107: do you need this as here and above?

Tolik170 (approved):

src/containers/offer-page/chat-dialog-window/ChatDialogWindow.tsx
TSlashDreamy: > do you need this as here and above?

Actually, yes, I haven't found any other way to transfer information between browser tabs other than localstorage. The information is deleted from localstorage, as soon as the user has been redirected to the desired chat

TSlashDreamy: > do you need this as here and above?

yup, without it, ts error appears

Mav-Ivan (approved):

mxrcury (approved):

maarekua (approved):

PR #1116, opened 2023-09-07
Search by messages in chat

ArturBekhDEV:
https://github.com/ita-social-projects/SpaceToStudy-Client/assets/102412173/17c49ddf-5f5c-45f8-b9e8-bc5494593d18

image


src/components/message/Message.styles.ts
```diff
@@ -13,6 +13,16 @@ export const styles = {
...
   },
+
+  findMessageCard: {
+    backgroundColor: 'primary.700',
+    borderRadius: '10px',
+    padding: '2px 4px',
+    display: 'inline',
+    color: 'white',
+    typography: TypographyVariantEnum.Body1,
+    p: '8px 16px'
```
Tolik170, line 24: also, you sett padding above

src/components/message/Message.tsx
```diff
@@ -46,11 +52,27 @@ const Message: FC<MessageProps> = ({ message, prevMessage, sx = {} }) => {
     e.stopPropagation()
   }
 
+  const messageRef = useRef<HTMLDivElement>(null)
+  const isTextFiltered = filteredMessages.some(
+    (filteredMessage) => filteredMessage.toLowerCase() === text.toLowerCase()
+  )
+  useEffect(() => {
+    if (filteredIndex >= 0 && filteredIndex < filteredMessages.length) {
```
Tolik170, line 60: maybe check that filteredIndex > 0 instead of filteredIndex >= 0?

src/components/message/Message.tsx
```diff
@@ -23,18 +23,24 @@ interface MessageProps {
   sx?: {
     avatar?: SxProps
   }
+  filteredMessages: string[]
```
Yurenko, line 26: Add it as not important because we will see warning

src/components/search-by-message/SearchByMessage.tsx
```diff
@@ -1,31 +1,62 @@
...
+}) => {
   const { t } = useTranslation()
   const [search, setSearch] = useState<string>('')
+  const [findMessage, setFindMessage] = useState<string[]>([])
 
+  const debouncedOnFilteredMessagesChange = useDebounce(
+    (filteredMessages: string[]) => {
+      onFilteredMessagesChange(filteredMessages)
+    },
+    1500
```
Yurenko, line 30: I guess this delay is too long, in my opinion, 500 will be enought.

src/components/message/Message.styles.ts
ArturBekhDEV: fixed

src/components/message/Message.tsx
ArturBekhDEV: In this case the index can be 0, because we are searching in an array

src/components/message/Message.tsx
ArturBekhDEV: fixed

src/components/search-by-message/SearchByMessage.tsx
ArturBekhDEV: yep. i think the same, right now its .5 sec

Tolik170 (approved):

TSlashDreamy (approved):

src/components/search-by-message/SearchByMessage.tsx
```diff
@@ -1,34 +1,69 @@
...
 
 import InputWithIcon from '~/components/input-with-icon/InputWithIcon'
 import IconsWithCounter from '~/components/icons-with-counter/IconsWithCounter'
+import { useDebounce } from '~/hooks/use-debounce'
 
 import { styles } from '~/components/search-by-message/SearchByMessage.styles'
 
 interface SearchByMessageProps {
-  maxValue: number
+  messages: { text: string }[]
```
mxrcury, line 12: maybe move this object type to `~/types`

mxrcury (approved):

Yurenko (approved):

maarekua (approved):

PR #1230, opened 2023-10-05
Unit tests for AdminPortal container

### LINKED ISSUE #1190: (SP:1) Write unit test for AdminPortal container
image

OlyaKorchan:
Created tests for AdminPortal container

1) Before 
image

2) After
Знімок екрана 2023-10-05 о 10 57 45

image

Closed #1190

### DIFF
--- a/tests/unit/containers/admin-portal/AdminPortal.spec.jsx
+++ b/tests/unit/containers/admin-portal/AdminPortal.spec.jsx
@@ -1,0 +1,54 @@
+import { configureStore } from '@reduxjs/toolkit'
+import reducer from '~/redux/reducer'
+import {
+  createMemoryRouter,
+  createRoutesFromElements,
+  RouterProvider
+} from 'react-router-dom'
+import { routerConfig } from '~/router/router'
+import { render, screen, waitFor } from '@testing-library/react'
+import { Provider } from 'react-redux'
+import { ThemeProvider } from '@mui/material/styles'
+import { theme } from '~/styles/app-theme/custom-mui.styles'
+import { beforeEach, expect } from 'vitest'
+
+const renderWithRouter = (initialEntries, preloadedState) => {
+  const store = configureStore({
+    reducer: { appMain: reducer },
+    preloadedState
+  })
+  const router = createMemoryRouter(createRoutesFromElements(routerConfig), {
+    initialEntries: [initialEntries]
+  })
+
+  render(
+    <Provider store={store}>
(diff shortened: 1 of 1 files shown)


tests/unit/containers/admin-portal/AdminPortal.spec.jsx
```diff
@@ -0,0 +1,53 @@
...
+const renderAdminPortalRouter = (initialEntries, preloadedState) => {
+  const store = configureStore({
+    reducer: { appMain: reducer },
+    preloadedState
+  })
+  const router = createMemoryRouter(createRoutesFromElements(routerConfig), {
+    initialEntries: [initialEntries]
+  })
+
+  render(
```
mxrcury, line 26: there is no need to add wrapping all these providers by yourself, we already have util [renderWithProviders](https://github.com/ita-social-projects/SpaceToStudy-Client/blob/82cefb038667db2ce8bfd941144bc5ce59e0ddab/tests/test-utils.jsx#L16) for that

OlyaKorchan: I have tried to use renderWithProviders and passed preloadedState with userRole - admin and initialEntries - '/admin', but in such case only empty div is rendered after AdminNavBar. Outlet component depends on routes and Suspence fallback is not working since Outlet is not loading. That is why I decided to add wrapping. 

Знімок екрана 2023-10-05 о 15 23 36

Знімок екрана 2023-10-05 о 15 22 55

abalanovsky (approved):

mxrcury (approved):

tests/unit/containers/admin-portal/AdminPortal.spec.jsx
```diff
@@ -0,0 +1,53 @@
...
+    <Provider store={store}>
+      <ThemeProvider theme={theme}>
+        <RouterProvider router={router}>
```
Tolik170, line 29: try to use renderWithProviders instead of these providers

OlyaKorchan: I have tried to use renderWithProviders and passed preloadedState with userRole - admin and initialEntries - '/admin', but in such case only empty div is rendered after AdminNavBar. Outlet component depends on routes and Suspence fallback is not working since Outlet is not loading. That is why I decided to add wrapping.

image

image

maarekua (approved):

PR #1249, opened 2023-10-08
Added test for FilterSelector

mynotdoing:
Added test for FilterSelector

image


tests/unit/components/filter-selector/FilterSelector.spec.jsx
```diff
@@ -0,0 +1,46 @@
...
+vi.mock('simplebar-react', () => {
+  return {
+    __esModule: true,
+    default: function MockedSimpleBar(props) {
+      return <div>{props.children}</div>
+    }
+  }
+})
+
+vi.mock('react-i18next', () => ({
```
Tolik170, line 14: I guess you don`t need this mock, you can find element using screen.getByText('common.noItems') and then check if this element is in document with .toBeInTheDocument()

tests/unit/components/filter-selector/FilterSelector.spec.jsx
```diff
@@ -0,0 +1,46 @@
...
+    t: (key) => {
+      const translations = {
+        'common.noItems': 'No items found'
+      }
+      return translations[key] || key // Return the custom translation or the key itself
+    }
+  })
+}))
+describe('FilterSelector', () => {
+  const selectedItems = []
```
Tolik170, line 25: unnecessary variable

tests/unit/components/filter-selector/FilterSelector.spec.jsx
```diff
@@ -0,0 +1,46 @@
...
+  test('FilterSelector component renders correctly', () => {
+    render(<FilterSelector selectedItems={selectedItems} />)
+    const el = screen.getByRole('button')
+    expect(el).toBeInTheDocument()
+  })
+
+  test('renders no items after clicking', async () => {
+    render(<FilterSelector selectedItems={[]} />)
```
Tolik170, line 33: if you render this component with the same props for each test, better move this render into forEach block and remove from tests

tests/unit/components/filter-selector/FilterSelector.spec.jsx
```diff
@@ -0,0 +1,46 @@
...
+    const buttonEl = screen.getByRole('button')
+    await user.click(buttonEl)
```
Tolik170, line 35: try to use fireEvent for click operations and maybe without using await

tests/unit/components/filter-selector/FilterSelector.spec.jsx
```diff
@@ -0,0 +1,46 @@
...
+    const noItemsText = screen.getByRole('presentation')
+    expect(noItemsText).toHaveTextContent('No items found')
+  })
+  test('icon should be in the component', async () => {
+    render(<FilterSelector selectedItems={[]} />)
+    const buttonEl = screen.getByRole('button')
+    await user.click(buttonEl)
+    const inputEl = screen.getByRole('textbox')
+    expect(inputEl).toBeInTheDocument()
```
Tolik170, line 44: add space between variable declarations, events and expects in tests

tests/unit/components/filter-selector/FilterSelector.spec.jsx
mynotdoing: without this mock the test passed but with an error Error: Not implemented: window.computedStyle(elt, pseudoElt). 
This mock allows me to avoid the error. What do you suggest in this case?

PR #1250, opened 2023-10-08
fixed placeholder for tutor's page

mynotdoing:
fixed placeholder for tutor's page
When a user is logged in as a Tutor, the placeholder on main page should correspond the mockup
Also added UA
image


src/constants/translations/en/tutor-home-page.json
```diff
@@ -4,7 +4,7 @@
         "tutor": "A Global Network of Students"
       },
       "description": "A Space2Study platform is a place where you can learn from the best. Type what would you like to teach and find students for that.",
-      "label": "What would you like to teach?",
+      "label": "Who or what do you want to teach?",
```
Tolik170, line 7: also, ua

Issue #1255, opened 2023-10-09
[My resources page] Duplication of elements when adding attachments


TSlashDreamy:
**Description:** Attachments are duplicated when we try to add a new one after the previous one.

**Environment:** Windows 11 x64, Firefox Developer Edition 119.0b6.

**Reproducible:** always.

**Preconditions:**
The https://s2s-front-stage.azurewebsites.net/ site is opened.
The user is logged in as a Tutor.

**Steps to reproduce:**
1. Click "My Resources" in the navigation bar.
2. Select "Attachments" tab.
3. Click "Add attachment" button, and add any file you want.
4. Repeat step 3.

**Actual result:**
Previous item was duplicated, when we added new one.

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/45912519/3aceaf52-4cf7-4aa7-8c88-30832b9d46ba

**Expected result:**
We are expecting just to add new item without duplicating previous one.

![зображення](https://github.com/ita-social-projects/SpaceToStudy-Client/assets/45912519/bbdc8585-2780-47ac-8417-4f36709d4a88)

PR #1313, opened 2023-10-16
Fixed attachments duplication on upload

abalanovsky:
**Before**:

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/108689551/35da6865-19c7-48a8-9c3d-c11bf3193d89

**After**:

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/108689551/352ac298-c35c-43b0-a63e-ac87b7f315b0

issue #1255

### DIFF
--- a/src/containers/add-documents/AddDocuments.tsx
+++ b/src/containers/add-documents/AddDocuments.tsx
@@ -21,7 +21,7 @@ interface AddDocumentsProps {
     button?: SxProps
   }
   icon?: ReactElement
-  keepPreviousFiles?: boolean
+  removePreviousFiles?: boolean
 }
 
 const AddDocuments: FC<AddDocumentsProps> = ({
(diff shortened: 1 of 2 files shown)


Mav-Ivan (approved):

MariaAfonina (approved):

dmtrth25 (approved):

src/containers/add-documents/AddDocuments.tsx
```diff
@@ -29,7 +30,8 @@ const AddDocuments: FC<AddDocumentsProps> = ({
   buttonText,
   variant,
   sx = {},
-  icon
+  icon,
+  keepPreviousFiles = true
```
maarekua, line 34: It is better to set the props in that way so the initial value will be false. Then you can be sure it will not affect old code.

like removePreviousFiles = false

maarekua (changes requested):

Tolik170 (approved):

TSlashDreamy (approved):

maarekua (approved):

PR #1409, opened 2023-11-15
removed access token from local storage

Tolik170:
Removed access token from local storage, now it will be stored in httpOnly cookie

### DIFF
--- a/src/types/user/user-interfaces/user.interfaces.ts
+++ b/src/types/user/user-interfaces/user.interfaces.ts
@@ -11,7 +11,7 @@ import {
 } from '~/types'
 
 export interface LocalStorage {
-  accessToken?: string
+  'emoji-mart.last'?: string
 }
 
 export interface GetUsersParams extends RequestParams {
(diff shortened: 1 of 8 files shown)


TSlashDreamy (approved):

OlyaKorchan (approved):

dmtrth25:
For better understanding you can add some description of where the token stored, screenshots etc.

MariaAfonina (approved):

abalanovsky (approved):

dmtrth25 (approved):

src/types/user/user-interfaces/user.interfaces.ts
```diff
@@ -11,7 +11,7 @@ import {
 } from '~/types'
 
 export interface LocalStorage {
-  accessToken?: string
+  'emoji-mart.last'?: string
```
maarekua, line 14: what is that for?

Tolik170: this interface represents the expected structure of the data stored in the local storage. 'emoji-mart.last' is the item that we store in our local storage

We use it in local-storage-service:
JSON.parse(localStorage.getItem('s2s') || 'null') as LocalStorage | null

maarekua (approved):

PR #1455, opened 2023-11-29
Added unique key to section in new course

OlyaKorchan:
If you delete the first section and then create a new one, the two last sections had the same id. 

Before 
Знімок екрана 2023-11-29 о 10 38 51

After 

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/17857767/498535fc-ab5b-489f-bfdc-f9a0c1ee3610

PR #1465, opened 2023-11-30
Created save logic for course

MariaAfonina:
https://github.com/ita-social-projects/SpaceToStudy-Client/assets/104988390/32c327c4-1c05-4657-b6fe-b1c32af14590

### DIFF
--- a/src/containers/course-section/CourseSectionContainer.tsx
+++ b/src/containers/course-section/CourseSectionContainer.tsx
@@ -154,9 +154,9 @@ const CourseSectionContainer: FC<SectionProps> = ({
       ...sectionValue,
       title: titleInput,
       description: descriptionInput,
-      lessons: lessons,
-      quizzes: quizzes,
-      attachments: attachments
+      lessons,
+      quizzes,
+      attachments
     }
 
     const newSectionsArray = [
--- a/src/containers/course-section/CourseSectionContainer.tsx
+++ b/src/containers/course-section/CourseSectionContainer.tsx
@@ -144,6 +144,30 @@ const CourseSectionContainer: FC<SectionProps> = ({
     setDescriptionInput(event.target.value)
   }
 
+  useEffect(() => {
+    const sectionIndex = sections.findIndex(
+      (section) => section.id === sectionData.id
+    )
+    let sectionValue = sections[sectionIndex]
+
+    sectionValue = {
+      ...sectionValue,
+      title: titleInput,
+      description: descriptionInput,
+      lessons: lessons,
+      quizzes: quizzes,
+      attachments: attachments
+    }
+
+    const newSectionsArray = [
+      ...sections.slice(0, sectionIndex),
+      sectionValue,
+      ...sections.slice(sectionIndex + 1, sections.length)
+    ]
+    setSectionsItems(newSectionsArray)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [titleInput, descriptionInput, lessons, quizzes, attachments])
+
   const handleAddResources = <T extends CourseResources>(
     resources: T[],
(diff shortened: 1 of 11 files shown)


src/containers/course-section/CourseSectionContainer.tsx
```diff
@@ -144,6 +144,30 @@ const CourseSectionContainer: FC<SectionProps> = ({
...
+    )
+    let sectionValue = sections[sectionIndex]
+
+    sectionValue = {
+      ...sectionValue,
+      title: titleInput,
+      description: descriptionInput,
+      lessons: lessons,
+      quizzes: quizzes,
+      attachments: attachments
```
TSlashDreamy, line 159: if key and value are the same you can write just a value
```js
sectionValue = {
      ...sectionValue,
      title: titleInput,
      description: descriptionInput,
      lessons
      quizzes,
      attachments
      }
```

src/containers/course-section/CourseSectionContainer.tsx
```diff
@@ -144,6 +144,30 @@ const CourseSectionContainer: FC<SectionProps> = ({
...
+    }
+
+    const newSectionsArray = [
+      ...sections.slice(0, sectionIndex),
+      sectionValue,
+      ...sections.slice(sectionIndex + 1, sections.length)
+    ]
```
TSlashDreamy, line 166: Have you tested the performance of the code? Wouldn't that slow it down with 10 sections or more? Is there a problem with the sections getting mixed up when you make changes or is that all right?

TSlashDreamy (approved):

Mav-Ivan (approved):

src/containers/course-section/CourseSectionContainer.tsx
MariaAfonina: Yes, I tested it. There are no problems if we make changes because it moves changes to the sections simultaneously when we change. But yes, it slows down a little when it is 10 sections.

TSlashDreamy: > Yes, I tested it. There are no problems if we make changes because it moves changes to the sections simultaneously when we change. But yes, it slows down a little when it is 10 sections.

please create task for refactoring later

OlyaKorchan (approved):

src/pages/create-course/CreateCourse.tsx
```diff
@@ -42,7 +80,7 @@ const CreateCourse = () => {
 
   const createNewSection = () => {
     const newSectionData = { ...sectionInitialData }
-    newSectionData.id = data.sections.length
+    newSectionData.id = Date.now() - data.sections.length
```
abalanovsky, line 83: Why do we need to assign an id?

maarekua (approved):

src/pages/create-course/CreateCourse.tsx
MariaAfonina: To know which sections were changed and for drop and drag.

dmtrth25: I suppose `newSectionData.id = Date.now()` will be sufficient for `newSectionData.id` or use data of sections length ? 
why are you subtracting the `data.sections.length` for id ?
```suggestion
    newSectionData.id = Date.now()
```

dmtrth25 (commented):
Preview with video link is downloaded to my local pc. fix the description

src/pages/create-course/CreateCourse.tsx
MariaAfonina: This line was created from this PR https://github.com/ita-social-projects/SpaceToStudy-Client/pull/1455 to not wait for merging. You can discuss this there. But I think you're right, maybe we can you without data.sections.length.

dmtrth25 (approved):

PR #1566, opened 2024-01-18
Created cooperation notes container

dmtrth25:
Implemented notes container with animation and Drawer

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/56305508/fd28b267-44aa-4554-afcd-d757a63ed435

**Resolved: #1560**

Issue #1555, opened 2024-01-15
(SP: 2) Create conditional menu Open/Open manually


dmtrth25:
Should render icon and text with conditional select option
Should change availability of activities after change

**Blocked** by #1556

<img width="615" alt="image" src="https://github.com/ita-social-projects/SpaceToStudy-Client/assets/56305508/56be2e58-d931-4527-a777-bebe4e73fe87">

<img width="627" alt="image" src="https://github.com/ita-social-projects/SpaceToStudy-Client/assets/56305508/4857d7ce-c70a-4372-99dc-ac23e35ee5c1">

![image](https://github.com/ita-social-projects/SpaceToStudy-Client/assets/17857767/5465523f-b782-4afd-820b-daf69e10213b)

PR #1576, opened 2024-01-24
Implemented resource availability 

abalanovsky:
The selection of resource availability is implemented

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/108689551/245e5b18-6476-4bb1-99cb-4bf3c3685a22

#1555 #1566


ArtemHolikov (approved):

src/containers/cooperation-details/cooperation-from-scratch/CooperationFromScratch.tsx
```diff
@@ -26,25 +26,46 @@ const CooperationFromScratch = () => {
...
   return (
-    <Box sx={styles.root}>
+    <Box data-testid='coop-from-scratch' sx={styles.root}>
       <Box sx={styles.publishBlock}>
         <Box>
           <Box sx={styles.lockBlock}>
-            <LockOpenTwoToneIcon sx={styles.icon} />
+            <Box
+              alt='rescource icon'
+              component='img'
```
Tolik170, line 35: ComponentEnum

src/containers/cooperation-details/cooperation-from-scratch/CooperationFromScratch.tsx
```diff
@@ -26,25 +26,46 @@ const CooperationFromScratch = () => {
...
+              src={
+                resourceAvailability === ResourcesAvailability.openAll
+                  ? openIcon
+                  : closeIcon
```
Tolik170, line 39: maybe is better to move this check outside, into a variable

src/containers/cooperation-details/cooperation-from-scratch/CooperationFromScratch.tsx
```diff
@@ -26,25 +26,46 @@ const CooperationFromScratch = () => {
...
+              }
+            />
             <Typography sx={styles.lockTitle}>
               {t('cooperationDetailsPage.publish')}
+              {t(
+                `cooperationDetailsPage.select.${
+                  resourceAvailability === ResourcesAvailability.openAll
+                    ? 'openAll'
+                    : 'openManually'
```
Tolik170, line 48: use enum for these values

src/containers/cooperation-details/cooperation-from-scratch/CooperationFromScratch.tsx
```diff
@@ -26,25 +26,46 @@ const CooperationFromScratch = () => {
...
+              )}
             </Typography>
           </Box>
           <Typography sx={styles.lockSubtitle}>
-            {t('cooperationDetailsPage.allResources')}
+            {t(
+              `cooperationDetailsPage.${
+                resourceAvailability === ResourcesAvailability.openAll
+                  ? 'allResources'
+                  : 'manuallyResources'
```
Tolik170, line 58: same here, use enum

src/containers/course-section/resource-item/ResourceItem.constants.ts
```diff
@@ -0,0 +1,16 @@
...
+    value: ResourceAvailabilityStatus.open,
+    title: 'Open'
+  },
+  {
+    value: ResourceAvailabilityStatus.openFrom,
+    title: 'Open from'
+  },
+  {
+    value: ResourceAvailabilityStatus.closed,
+    title: 'Closed'
```
Tolik170, line 14: Enum for title

src/containers/course-section/resource-item/ResourceItem.tsx
```diff
@@ -1,23 +1,58 @@
...
   resource: CourseResources
   deleteResource: (resource: CourseResources) => void
+  setResourceAvailability: (
+    id: string,
+    availability: ResourceAvailabilityStatus,
+    openFromDate: string | null
+  ) => void
+}
+
+const availabilityIcons = {
```
Tolik170, line 38: will we use these icons in other files in the future? If yes, makes sense to move them outside and import

src/containers/course-section/resource-item/ResourceItem.tsx
```diff
@@ -26,16 +61,80 @@ const ResourceItem: FC<ResourceItemProps> = ({ resource, deleteResource }) => {
...
+            <Box
+              alt='resource icon'
+              component='img'
+              src={availabilityIcons[resourceAvailabilityStatus]}
+              sx={styles.availabilityIcon}
+            />
+            <DatePicker
+              disablePast
+              inputFormat={'MMM d, yyyy'}
+              label='Opening date'
```
Tolik170, line 108: translation for label

src/containers/course-section/resources-list/ResourcesList.tsx
```diff
@@ -56,6 +56,21 @@ const ResourcesList: FC<ResourcesListProps> = ({
     setResources(reorderedItems)
   }
 
+  const setResourceAvailability = (
+    id: string,
+    availability: ResourceAvailabilityStatus,
+    openFromDate?: string | null
+  ) => {
+    const resources = [...items]
+    const resource = resources.find((item) => item._id === id)
```
Tolik170, line 65: you can use items.find and below use items in if statement, so resources variable is redundant, also, you can rename items prop to resources

src/types/my-cooperations/enums/myCooperations.enums.ts
```diff
@@ -0,0 +1,10 @@
+export enum ResourcesAvailability {
+  openAll = 'openAll',
+  openManually = 'openManually'
+}
+
+export enum ResourceAvailabilityStatus {
```
Tolik170, line 6: Add Enum to the end of each enum: 
ResourcesAvailabilityEnum,
ResourceAvailabilityStatusEnum

src/containers/course-section/resource-item/ResourceItem.tsx
```diff
@@ -26,16 +61,80 @@ const ResourceItem: FC<ResourceItemProps> = ({ resource, deleteResource }) => {
     }
   }
 
+  const resourceAvailabilityStatus =
+    resource.resourceAvailability ?? ResourceAvailabilityStatus.open
+  const displayDatePicker =
+    resourceAvailabilityStatus === ResourceAvailabilityStatus.openFrom
+
```
Tolik170, line 68: better use destructure for ResourceAvailabilityStatus

src/containers/course-section/resource-item/ResourceItem.constants.ts
abalanovsky: replaced with translation

src/containers/course-section/resource-item/ResourceItem.tsx
abalanovsky: Not expected

src/containers/course-section/resources-list/ResourcesList.tsx
abalanovsky: I can't directly change element of array in useState

OlyaKorchan (approved):

src/types/my-cooperations/enums/myCooperations.enums.ts
```diff
@@ -0,0 +1,10 @@
+export enum ResourcesAvailabilityEnum {
+  openAll = 'openAll',
+  openManually = 'openManually'
+}
```
dmtrth25, line 4: keep in mind convention for enum naming

image

src/types/course/types/course.types.ts
```diff
@@ -1,6 +1,13 @@
...
+  Quiz,
+  Attachment,
+  ResourceAvailabilityStatusEnum
+} from '~/types'
 
-export type CourseResources = Lesson | Quiz | Attachment
+type Resources = Lesson | Quiz | Attachment
 
-export type MyCorsesListProps = Omit<ItemsWithCount<Course>, 'count'>
+export type CourseResources = Resources & {
```
dmtrth25, line 10: here better use interface extends
like 
`interface CourseResources extends Resources` and add your fields

abalanovsky: It's type, actually

dmtrth25: > It's type, actually

I told you about interface

FYI
image

src/containers/course-section/resources-list/ResourcesList.tsx
Tolik170: you can use:
setResources([...resources])

and in other places where you are using resources you can just use items, so, you will avoid to create the resources variable

src/containers/course-section/resources-list/ResourcesList.tsx
```diff
@@ -56,6 +56,21 @@ const ResourcesList: FC<ResourcesListProps> = ({
...
+    id: string,
+    availability: ResourceAvailabilityStatusEnum,
+    openFromDate?: string | null
+  ) => {
+    const resources = [...items]
+    const resource = resources.find((item) => item._id === id)
+
+    if (resource) {
+      resource.resourceAvailability = availability
+      openFromDate !== undefined && (resource.openFromDate = openFromDate)
```
Tolik170, line 69: do you need to check `openFromDate !== undefined` ?
I think, `openFromDate && (resource.openFromDate = openFromDate)` should work

abalanovsky: Actually I do. `openFromDate` can be null and we need to distinguish between null and undefined. The solution you are suggesting will not set the null value for the date.

src/types/course/types/course.types.ts
abalanovsky: `Resources` is not a static type, so an interface cannot extend from it. 

telegram-cloud-photo-size-4-5881759469522764621-y


Extending interfaces of Lesson, Quiz and Attachment does not look good to me, since we will have to do that in three separate files and those interfaces are used not only for the Resources type. Therefore it makes sense to only apply these additional fields here.

src/containers/course-section/resources-list/ResourcesList.tsx
Tolik170: ok, but I don`t see the place where it can set undefined to openFromDate, as I see according to code and interface in ResourceItem where you passed this function, openFromDate can be null or string but not undefined, or I am wrong?

src/containers/course-section/resources-list/ResourcesList.tsx
Tolik170: ok, but I don`t see the place where it can set undefined to openFromDate, as I see according to code and interface in ResourceItem where you passed this function, openFromDate can be null or string but not undefined, or I am wrong?

src/containers/course-section/resources-list/ResourcesList.tsx
abalanovsky: I cannot get the required item using items.find because items is a prop and I cannot directly make changes to the found object. It will not update at all.

src/containers/course-section/resources-list/ResourcesList.tsx
abalanovsky: I did that in case we will need an opportunity to call the function without specifying date at all. But yeah, I don't think I use it in that way so I will remove the check.

abalanovsky: Done

Tolik170 (approved):

Mav-Ivan (approved):

maarekua (approved):

dmtrth25 (approved):

PR #1644, opened 2024-02-27
Feature/708/proxying local requests

abalanovsky:
Proxying requests have been added only locally.
Cookie issues resolved.

Знімок екрана 2024-02-27 о 13 31 16

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -19,7 +19,14 @@ export default defineConfig({
   server: {
     port: 3000,
     open: true,
-    host: true
+    host: true,
+    proxy: {
+      '/api': {
+        target: 'http://localhost:8080',
+        changeOrigin: true,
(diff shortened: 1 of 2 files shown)


vite.config.ts
```diff
@@ -19,7 +19,14 @@ export default defineConfig({
   server: {
     port: 3000,
     open: true,
-    host: true
+    host: true,
+    proxy: {
+      '/api': {
+        target: 'http://localhost:8080',
```
Tolik170, line 25: we have a variable in env for api path, can you use it here?

abalanovsky: > we have a variable in env for api path, can you use it here?

`.env` will be changed after this pr so I can just create an extra variable for `.env` but not sure if that makes sense

Tolik170 (approved):

Mav-Ivan (approved):

Issue #409, opened 2023-01-27
(SP: 3) Implement block 'About Tutor' on 'Tutor's profile'


KhrystynaPavlikovska:
Implement block 'About Tutor' on 'Tutor's profile' according to design on Figma.

![image](https://user-images.githubusercontent.com/34419998/215153503-672dbe1f-c2f0-470a-bc50-004a0d54eca0.png)

PR #1662, opened 2024-03-14
Created About Tutor block for my profile

dmtrth25:
Implemented accordion `About the Tutor` block for my profile page

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/56305508/9f4478f2-fcba-4f4d-9b2d-b4eed62c1309

**Resolved: #409**


src/constants/translations/en/tutor-profile-page.json
```diff
@@ -20,6 +20,19 @@
   "videoPresentation": {
     "title": "Video Presentation"
   },
+  "aboutTutor": {
```
Tolik170, line 23: There is no reason to add this text to translation, in future tutor will paste what he wants there

dmtrth25: > There is no reason to add this text to translation, in future tutor will paste what he wants there

Sure. but for now it a good place to pass text. When we'll work with this feature then we can replace logic. For now it's mocked text data.
Suggest leave it only inside in the component without translations ?

Mav-Ivan (approved):

abalanovsky (approved):

Tolik170 (approved):

PR #1703, opened 2024-04-02
Fixed alignment and style of error message in add photo step of the stepper

GeorgeShvab:
Alignment of error message in the stepper was not properly aligned.  Also, colour of the error message was inconsistent with the error message in previous step (subject step). Fixed for both tutor and student.

Before fix:
318134154-9ae9879c-af05-43d8-b514-9a8c293ba8eb

After fix:
erro-message-fix

### DIFF
--- a/src/components/file-uploader/FileUploader.styles.ts
+++ b/src/components/file-uploader/FileUploader.styles.ts
@@ -29,10 +29,5 @@ export const styles = {
   fileSize: {
     mt: '10px',
     typography: TypographyVariantEnum.Body2
-  },
-  error: {
-    color: 'error',
-    ml: 1,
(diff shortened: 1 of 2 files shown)


maarekua (approved):

abalanovsky (approved):

iva-stasia (commented):
Maybe it's better to change the error style directly in the`FileUploader.styles.ts` file so that the error style is the same everywhere this component is used? What do you think?

Tolik170 (approved):

D1nviz (approved):

Mav-Ivan (approved):

iva-stasia (approved):

PR #1706, opened 2024-04-02
Bugfix/1695/attachment error

asya982:
- Separate logic in validation for possible futher partial validating
- Added validation for maxFileNameSize
- Wrote tests for files util
- Added and fixed values in translations

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/105641724/4f3cbed9-3c0f-4c8d-a52f-a7f3e6a352f7


maarekua (approved):

src/constants/translations/en/common.json
```diff
@@ -38,8 +38,9 @@
   "megabytes": "MB",
   "uploadNewFile": "Upload new file",
   "typeError": "Wrong file type. Only .pdf/.jpg/.jpeg files are allowed.",
-  "allFilesSizeError": "All files size error",
+  "allFilesSizeError": "Size of all files cannot be more than 20 MB",
   "fileSizeError": "Size for one file cannot be more than 5 MB.",
+  "fileNameError":"File name cannot be longer than 55 symbol",
```
mhevyk, line 43: Maybe "File name cannot be longer than 55 characters" with be better description of error?

asya982: makes sense, even though user won't ever count his file name length😄
updated✅

src/containers/add-documents/AddDocuments.constants.ts
```diff
@@ -7,5 +7,7 @@ export const validationData: AddDocuments = {
   fileSizeError: 'common.fileSizeError',
   allFilesSizeError: 'common.allFilesSizeError',
   typeError: 'common.typeError',
-  maxQuantityFiles: 7
+  maxQuantityFiles: 7,
```
mhevyk, line 10: 1. Seems like we limit number of files uploaded in one attempt, but check is not used inside validations
2. I noticed, that if I upload multiple files, they upload in random order (I believe it depends on file size or filename)
3. It would be great user experience if we sort these files in some way

Maybe it can be moved into new issue or fixed here

asya982: ```maxQuantityFiles``` is used in hook ```use-upload``` checks, it is needed for slicing all files, which are user select through the input, and than ```filesValidaion``` checks files. So it's basically not possible to get more than 7 files

about sorting, i think i can look what can we do here

mhevyk: In my opinion it is bad user experience, because if user wants to upload 20 files (for example his or her folder with resources) he actually uploads only 7 files and it is confusing and not user-friendly. I think we should inform user about it in validation

asya982: added error for max quantity, as for sorting it should be backend task to add sort by creation when fetching attachments

mhevyk: It shows error "Maximum allowed to add 7 photos at once" for the first time, but second time when i attempt to load more than 7 files at once i don`t get an error and files are uploaded. Maybe something is not reset properly

asya982: will fix it in another issue

t-valentyna (approved):

D1nviz (approved):

tests/unit/utils/files.spec.jsx
```diff
@@ -0,0 +1,112 @@
...
+    type: 'image/jpeg',
+    webkitRelativePath: ''
+  },
+  {
+    lastModified: 1676387080931,
+    lastModifiedDate: new Date(),
+    name: 'wwwwwwwwwwwwwwwww.jpg',
+    size: 5_000_000,
+    type: 'image/jpeg',
+    webkitRelativePath: ''
```
Tolik170, line 88: instead of repeating same fields with value for each file, you can create an initial file and then use spread operator

PR #1709, opened 2024-04-02
Turn the category filter into a button on the My Resources page

D1nviz:
### Renamed, decomposed this component to AppMenuButton, AppSelectButton. Also i wrote tests

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/118522791/2eb9ac29-3051-4718-b8a3-37b4f9890cb7

### Test coverages
image
image

### DIFF
--- a/src/components/app-button-menu/AppButtonMenu.tsx
+++ b/src/components/app-button-menu/AppButtonMenu.tsx
@@ -1,15 +1,11 @@
-import { useState, ChangeEvent, useMemo, useEffect } from 'react'
+import { useState, useMemo, useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
 import SimpleBar from 'simplebar-react'
 import { SxProps } from '@mui/material'
 import Box from '@mui/material/Box'
 import { MenuProps } from '@mui/material/Menu'
-import MenuItem from '@mui/material/MenuItem'
 import { PopoverOrigin } from '@mui/material/Popover'
-import Divider from '@mui/material/Divider'
-import Checkbox from '@mui/material/Checkbox'
 import Typography from '@mui/material/Typography'
-import ClearIcon from '@mui/icons-material/Clear'
 import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline'
 import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown'
 
@@ -17,7 +13,8 @@ import useAxios from '~/hooks/use-axios'
 import useMenu from '~/hooks/use-menu'
 import Loader from '~/components/loader/Loader'
 import AppButton from '~/components/app-button/AppButton'
-import InputWithIcon from '~/components/input-with-icon/InputWithIcon'
+import AppMenuButton from '~/components/app-menu-button/AppMenuButton'
+import AppSelectButton from '~/components/app-select-button/AppSelectButton'
 
 import { defaultResponses } from '~/constants'
 import { spliceSx } from '~/utils/helper-functions'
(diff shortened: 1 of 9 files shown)


src/components/filter-selector/FilterSelector.styles.ts
```diff
@@ -11,7 +11,6 @@ export const styles = {
     px: '12px',
```
maarekua, line 11: I would like to suggest renaming this component. As it could be used not only as filter.

maarekua (changes requested):

src/components/filter-selector/FilterSelector.styles.ts
D1nviz: So, in your opinion, which name would be more flexible? AppSelector, for example?

src/components/filter-selector/FilterSelector.tsx
```diff
@@ -141,17 +140,17 @@ const FilterSelector = <T extends Pick<CategoryNameInterface, '_id'>>({
...
 
   return (
-    <Box id='menu-filter' sx={spliceSx(styles.root, customSx?.root)}>
-      <Typography sx={styles.text}>{title}:</Typography>
-      <Typography sx={styles.chosenFilters}>{chosenFiltersText}</Typography>
-      <IconButton
-        disableRipple
+    <>
+      <AppButton
+        id='menu-filter'
```
mhevyk, line 145: What is "id" prop used for here?

t-valentyna (approved):

src/components/filter-selector/FilterSelector.tsx
mhevyk: Maybe this can be good use case for useId hook, because if we try to reuse this component in other places something can go wrong because id will not be unique

D1nviz: Thank you. Replaced id with ref, deleted document.getById

src/components/multi-select/MultiSelect.tsx
```diff
@@ -54,6 +53,7 @@ const FilterSelector = <T extends Pick<CategoryNameInterface, '_id'>>({
   const [menuAnchor, setMenuAnchor] = useState<null | HTMLElement>(null)
   const [inputValue, setInputValue] = useState<string>('')
   const [selectedNames, setSelectedNames] = useState<string[]>([])
+  const AppButtonRef = useRef<HTMLButtonElement | null>(null)
```
mhevyk, line 56: I think `appButtonRef ` should be better name, because we use CamelCase for ref variables

iva-stasia (changes requested):
We don't need so many changes to achieve the desired result.
Please take a look at the documentation for the [MUI menu](https://mui.com/material-ui/react-menu/#basic-menu) and pay attention to how the `anchorEl` setting is handled.
 
This way, you do not need to modify the `AppButton` component and so on as well as use `ref` or `id`.
You can also consider using our custom `use-menu` hook.

src/components/filter-selector/FilterSelector.styles.ts
maarekua: You tell me :) it is a button, with menu. Looks like AppButtonMenu. And also we can decompose it to,

AppMenuButton that would have children for menu section

And

AppSelectButton as AppMenuButton with select container for menu

src/components/multi-select/MultiSelect.tsx
D1nviz: ` const handleInputChange = (e: ChangeEvent<HTMLInputElement>) => {
    setInputValue(e.target.value)
  }`
  
 Ref replaced by event.target

src/components/filter-selector/FilterSelector.styles.ts
D1nviz: I used the useMenu hook to render the menu in this component. I think its good way to make this component more readable and more flexible. And also we use render JSX from hook, not in this component. I made small refactors, that made this component more clean

src/components/app-button-menu/AppButtonMenu.tsx
```diff
@@ -49,11 +49,11 @@ const FilterSelector = <T extends Pick<CategoryNameInterface, '_id'>>({
   showNoneProperty = false,
   customSx,
   ...props
-}: FilterSelectorProps<T>) => {
+}: AppButtonMenuProps<T>) => {
   const { t } = useTranslation()
-  const [menuAnchor, setMenuAnchor] = useState<null | HTMLElement>(null)
```
mhevyk, line 54: Is this state used anywhere?

mhevyk (approved):

t-valentyna (approved):

byavd (approved):

iva-stasia (approved):

maarekua (approved):

PR #1728, opened 2024-04-03
Converted useUpload hook to typescript

mhevyk:
Converted useUpload hook to typescript


GeorgeShvab (approved):

src/containers/add-documents/AddDocuments.tsx
```diff
@@ -43,11 +43,11 @@ const AddDocuments: FC<AddDocumentsProps> = ({
         severity: snackbarVariants.error,
         message: documentsError
       })
-      setDocumentsError('')
+      setDocumentsError(undefined)
```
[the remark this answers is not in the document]
maarekua: Lets stay then with empty string, why do we need undefined?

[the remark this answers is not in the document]
mhevyk: Makes sense, fixed

src/components/file-uploader/FileUploader.tsx
```diff
@@ -17,16 +17,16 @@ import {
...
+  UploadFileEmitter,
   InputEnum,
   SizeEnum
 } from '~/types'
 
 interface FileUploaderProps {
   buttonText: string
-  emitter: ({ files, error }: Emitter) => void
+  emitter: UploadFileEmitter
   initialState: File[]
```
iva-stasia, line 28: If `initialState ` has a default value, could it be optional (similar to `initialError `)? It will be useful if we don't have an initial value.

iva-stasia (changes requested):

src/components/file-uploader/FileUploader.tsx
mhevyk: Thanks for the comment, I didn't take it into account this tiny possible typescript issue when performing this task. Already fixed

iva-stasia (approved):

src/containers/tutor-home-page/add-photo-step/AddPhotoStep.jsx
```diff
@@ -13,7 +13,7 @@ import { validationData } from './constants'
 
 const AddPhotoStep = ({ btnsBox, stepLabel }) => {
   const { isLaptopAndAbove, isTablet, isMobile } = useBreakpoints()
-  const [photoError, setPhotoError] = useState(null)
+  const [photoError, setPhotoError] = useState()
```
Tolik170, line 16: what was the purpose to make undefined as the default value instead of null?

mhevyk: I did this because `UploadFileEmitter` accepts optional error like following:
```jsx
type UploadFileEmitterArgs = {
  files: File[]
  error?: string
}

export type UploadFileEmitter = (args: UploadFileEmitterArgs) => void
```
I think it will be finally close all inaccuracies after we convert `AddPhotoStep` component and other components related to file uploading to typescript.

mhevyk: Speaking about `error ?? ''` inside `AddDocuments.tsx`, I have two options to fix type error with current implementation as shown below:

image

Please tell me which one should I pick

I think the same issue will be inside `AddPhotoStep.jsx`, but it is not converted to typescript yet.

D1nviz (approved):

src/containers/tutor-home-page/add-photo-step/AddPhotoStep.jsx
Tolik170: I prefer to use the second option, but as I see, you changed an error to optional, and that`s why it throws an error, can you make the error not optional in Emmiter?

asya982 (approved):

src/containers/tutor-home-page/add-photo-step/AddPhotoStep.jsx
mhevyk: Made `Emitter` error optional

maarekua (approved):

byavd (approved):

Issue #1721, opened 2024-04-03
(SP: ) Adding styles to EditProfile container according to mockup


OliaHrytsak:
Add neсessary style changes for  EditProfile container according to mockup

Edit Profile  container
![image](https://github.com/ita-social-projects/SpaceToStudy-Client/assets/127191734/402e9044-c163-4c63-95cd-140c184decd7)

Mockup
![image](https://github.com/ita-social-projects/SpaceToStudy-Client/assets/127191734/589e1f03-e60e-4ef2-8b5d-2f4a1cadf682)

PR #1732, opened 2024-04-03
Styles for edit profile container #1721

OliaHrytsak:
-Added styles for EditProfile container accordance to mockup
-Added styles for SidebarMenu component into separate style file
There no mockup for mobile

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/127191734/c16c1fae-1742-4035-8474-651e63acf93c

Mockup

image

### DIFF
--- a/src/pages/edit-profile/EditProfile.styles.ts
+++ b/src/pages/edit-profile/EditProfile.styles.ts
@@ -1,6 +1,10 @@
 import { TypographyVariantEnum } from '~/types'
 
 export const styles = {
+  sectionsWrapper: {
+    pr: { md: '156px', lg: '156px' },
+    pl: { md: '156px', lg: '156px' }
+  },
   headerContainer: {
     display: 'flex',
     justifyContent: 'space-between',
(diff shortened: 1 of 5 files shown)


src/pages/edit-profile/EditProfile.styles.ts
```diff
@@ -1,13 +1,18 @@
+import palette from '~/styles/app-theme/app.pallete'
 import { TypographyVariantEnum } from '~/types'
 
 export const styles = {
+  sectionsWrapper: {
+    pr: { lg: '156px' },
+    pl: { lg: '156px' }
+  },
```
mhevyk, line 8: Maybe it would be better to use `px: { lg: '156px' }`?

src/pages/edit-profile/EditProfile.styles.ts
```diff
@@ -20,14 +25,24 @@ export const styles = {
...
+    justifyContent: 'space-between',
+    '& .MuiButtonBase-root:hover': {
+      backgroundColor: palette.basic.grey,
+      '& .MuiTypography-body1': {
+        color: palette.basic.black,
+        fontWeight: '600'
+      },
+      '& .MuiListItemIcon-root svg': {
+        color: palette.basic.black
+      }
```
mhevyk, line 42: These selectors look a bit confusing. `EditProfile` component uses `SidebarMenu` component, which is used only once in our app. Please consider adding this styles to button directly or at least move it from `mainContainer` to `sidebarMenu` style and make more speciffic selector to make style more modular

src/pages/edit-profile/EditProfile.styles.ts
OliaHrytsak: Thanks! I've fixed it

PR #1733, opened 2024-04-04
Fixed validation of signup form

GeorgeShvab:
I changed name validation pattern to match only values without spaces (Previously it was not handled on a client side, so server returned error in case of space in name or surname).

I also splited password validation pattern into two cases. One of them matches values with one alphabetic character and one numeric character, another one matches only valid values. I did because previously in both cases "Password must contain at least one alphabetic and one numeric character" error was returned. That is not descriptive in case of invalid character error (if there is a space in password, for example).

I also added new error message for invalid symbol in password.

Wrote tests for common validations.

**Before fix:**

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/62070431/5c5007cb-69fc-4dc5-9110-a159991cbd02

**After fix:**

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/62070431/8694780c-afe1-48a6-aa96-3a88b0e29b7e

### DIFF
--- a/src/constants/translations/en/common.json
+++ b/src/constants/translations/en/common.json
@@ -69,7 +69,8 @@
   "errorMessages": {
     "emailValid": "Email should be of the following format: “local-part@domain.com”",
     "passwordLength": "Password cannot be shorter than 8 and longer than 25 characters",
-    "passwordValid": "Password must contain at least one alphabetic and one numeric character",
+    "passwordAlphabeticAndNumeric": "Password must contain at least one alphabetic and one numeric character",
+    "passwordValidSymbols": "Password must contain only alphabetic, numeric and special characters",
     "emptyField": "This field cannot be empty",
     "nameLength": "This field cannot be longer than 30 characters",
     "nameAlphabeticOnly": "This field can contain alphabetic characters only",
--- a/src/utils/validations/common.ts
+++ b/src/utils/validations/common.ts
@@ -1,3 +1,5 @@
+import { validationPatterns } from '~/utils/validations/validations.constants'
+
 interface Validations {
   nameField: (value: string) => string
   numberField: (value: string) => string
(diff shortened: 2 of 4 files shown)


src/constants/translations/en/common.json
```diff
@@ -69,7 +69,8 @@
   "errorMessages": {
     "emailValid": "Email should be of the following format: “local-part@domain.com”",
     "passwordLength": "Password cannot be shorter than 8 and longer than 25 characters",
-    "passwordValid": "Password must contain at least one alphabetic and one numeric character",
+    "passwordAlphabeticAndNumeric": "Password must contain at least one alphabetic and one numeric character",
+    "passwordValidSymbols": "Password must contain only valid symbols",
```
iva-stasia, line 73: As a user, I can't know what is a 'valid symbol' here.
Could it be better to use something like `Password must contain only alphabetic and numeric characters`?

iva-stasia (changes requested):

iva-stasia:
Please don't forget to link the issue.

asya982 (approved):

iva-stasia (approved):

t-valentyna (approved):

src/constants/translations/en/common.json
```diff
@@ -69,7 +69,8 @@
   "errorMessages": {
     "emailValid": "Email should be of the following format: “local-part@domain.com”",
     "passwordLength": "Password cannot be shorter than 8 and longer than 25 characters",
-    "passwordValid": "Password must contain at least one alphabetic and one numeric character",
+    "passwordAlphabeticAndNumeric": "Password must contain at least one alphabetic and one numeric character",
+    "passwordValidSymbols": "Password must contain only alphabetic, numeric and special characters",
```
mhevyk, line 73: Did yout provide corresponding translations in ua language?

src/utils/validations/common.ts
```diff
@@ -10,7 +10,7 @@ const validations: Validations = {
     if (value.length > 30) {
       return 'common.errorMessages.nameLength'
     }
-    if (!RegExp(/^(?! )[a-zа-яєії ]+(?<! )$/i).test(value)) {
+    if (!RegExp(/^[a-zа-яєії]+$/i).test(value)) {
```
mhevyk, line 13: Is there a need to use RegExp constructor or we can omit it?

src/constants/translations/en/common.json
GeorgeShvab: Regarding translation, there are no corresponding ua translations for many messages. There is no even `errorMessages` object in `translations/ua/common.json`. So I thought it would be better to make separate pr with all missing ua translations.

src/utils/validations/common.ts
GeorgeShvab: You are right. I guess it would be even better to make a separate object with validation patterns.

src/constants/translations/en/common.json
mhevyk: Hope to see new pull request

mhevyk (approved):

maarekua (approved):

D1nviz (approved):

PR #1735, opened 2024-04-04
Fixed offer card title width

t-valentyna:
Fixed offer card title width.
Before:
image

After fixing:
image

### DIFF
--- a/src/containers/find-offer/offer-card-square/OfferCardSquare.styles.ts
+++ b/src/containers/find-offer/offer-card-square/OfferCardSquare.styles.ts
@@ -32,7 +32,8 @@ export const styles = {
     ...ellipsisTextStyle(2),
     typography: 'midTitle',
     fontWeight: 600,
-    color: 'primary.700'
+    color: 'primary.700',
+    wordBreak: 'break-word'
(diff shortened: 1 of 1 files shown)


Tolik170 (approved):

asya982 (approved):

iva-stasia (commented):
Your solution is good!
However, I'm not sure that such a different shift in content on the cards looks good.

A possible solution is to leave space for the second line in advance, but I'm not sure if this is better than the current result.
image

What do you think?

t-valentyna:
> Your solution is good! However, I'm not sure that such a different shift in content on the cards looks good.
> 
> A possible solution is to leave space for the second line in advance, but I'm not sure if this is better than the current result. image
> 
> What do you think?

It's a great proposal. Unfortunately, on the mockup we can see only cards with long titles, so I have no idea, what the designer's vision was about one-line titles. Maybe we could wait for details about the design before changing this look.

abalanovsky (approved):

iva-stasia (approved):

D1nviz (approved):

PR #1737, opened 2024-04-04
conditional rendering in carouseWithImage if description is undefined

asya982:
to prevent errors when description is undefined added additional check for item.description

### DIFF
--- a/src/containers/course-section/CourseSectionContainer.tsx
+++ b/src/containers/course-section/CourseSectionContainer.tsx
@@ -42,20 +42,12 @@ import {
   Quiz,
   Attachment,
   ResourcesTabsEnum as ResourcesTypes,
-  CourseResources
+  CourseResources,
+  CourseSectionHandlers
 } from '~/types'
 import { styles } from '~/containers/course-section/CourseSectionContainer.styles'
-import CourseSectionsList from '../course-sections-list/CourseSectionsList'
 
-interface SectionProps
-  extends Pick<
-    Parameters<typeof CourseSectionsList>[0],
-    | 'handleSectionInputChange'
-    | 'handleSectionNonInputChange'
-    | 'handleSectionResourcesOrder'
-    | 'titleText'
-    | 'setSectionsItems'
-  > {
+interface SectionProps extends CourseSectionHandlers {
   sectionData: CourseSection
   sections: CourseSection[]
 }
--- a/src/types/common/types/common.types.ts
+++ b/src/types/common/types/common.types.ts
@@ -45,9 +45,9 @@ export type UseFormEventHandler<Fields, Event> = (
   key: keyof Fields
 ) => (event: Event) => void
 
-type FormValidationHandler<TData, TValue> = (
-  value: TValue | string,
-  data: TData
+type FormValidationHandler<Data, Value> = (
+  value: Value | string,
+  data: Data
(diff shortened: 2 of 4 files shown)


src/components/carousel-with-image/CarouselWithImage.tsx
```diff
@@ -31,7 +31,9 @@ const CarouselWithImage: FC<CarouselWithImageProps> = ({ items }) => {
       />
       <Box sx={styles.text}>
         <Typography sx={styles.title}>{t(item.title)}</Typography>
-        <Typography sx={styles.description}>{t(item.description)}</Typography>
+        <Typography sx={styles.description}>
+          {t(item.description ?? '')}
```
maarekua, line 35: What would be translation of empty string?

maarekua (changes requested):

src/components/carousel-with-image/CarouselWithImage.tsx
asya982: solved

src/containers/course-section/CourseSectionContainer.tsx
```diff
@@ -45,26 +45,19 @@ import {
   CourseResources
 } from '~/types'
 import { styles } from '~/containers/course-section/CourseSectionContainer.styles'
+import CourseSectionsList from '../course-sections-list/CourseSectionsList'
```
Tolik170, line 48: use import alias

GeorgeShvab (approved):

src/containers/course-section/CourseSectionContainer.tsx
```diff
@@ -45,26 +45,19 @@ import {
...
 
-interface SectionProps {
+interface SectionProps
+  extends Pick<
+    Parameters<typeof CourseSectionsList>[0],
```
Tolik170, line 52: can you explain how ```
extends Pick<
    Parameters<typeof CourseSectionsList>[0] 
```should work and why you can`t just extend from another interface or define these properties here like before

File: src/types/common/types/common.types.ts [hunk 4]
```diff
@@ -29,3 +29,29 @@ export type Country = {
...
+export type FormNonInputValueChange<Value, Fields> = (
+  key: keyof Fields,
+  value: Value
+) => void
+
+export type UseFormEventHandler<Fields, Event> = (
+  key: keyof Fields
+) => (event: Event) => void
+
+type FormValidationHandler<TData, TValue> = (
```
Tolik170, line 48: If you didn`t use **T** character before generic name in other types, remove it here too

asya982: solved

src/containers/course-section/CourseSectionContainer.tsx
asya982: i moved this repeating types to common interface `CourseSectionHandlers` and extend both components from it

src/containers/course-section/CourseSectionContainer.tsx
asya982: deleted this line, because it's unused now

maarekua (approved):

D1nviz (approved):

mhevyk (approved):

Tolik170 (approved):

PR #1778, opened 2024-04-09
Fixed bug when find offer toolbar was not mapped correctly to search params

mhevyk:
State in toolbar is now synchronized with search params as shown in the video below (also works for both tutor's and student's requests):

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/109922101/63422792-15af-45d1-ad9c-57f805733394

The first thing, that caused problems was `updateFilterInQuery` method. There were several places where they were called consequently like following:

```jsx
updateFilterInQuery(value?._id ?? '', 'category')
updateFilterInQuery('', 'subject')
```

It caused problems, because browser url was changed twice (via `setSearchParams` call). As a result, the connection between history and state was lost. So I had to set all required search params per one action
To fix this bug I refactored method for updating filters as following:

```jsx
updateFiltersInQuery({ category: value?._id ?? '', subject: '' })
```

This method can accept as many filters as you want and it is type safe.
Also I deleted old method to avoid same issues in future.

Also I updated types and tests

### DIFF
--- a/src/hooks/use-filter-query.tsx
+++ b/src/hooks/use-filter-query.tsx
@@ -1,6 +1,6 @@
-import { useState, useCallback } from 'react'
+import { useCallback, useMemo } from 'react'
 import { useSearchParams } from 'react-router-dom'
-import { FindOffersFiltersActions } from '~/types'
+import { FindOffersFiltersActions, UpdateFiltersInQuery } from '~/types'
 import { parseQueryParams } from '~/utils/helper-functions'
 
 interface UseFilterQueryOptions<T> {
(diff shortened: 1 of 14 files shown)


t-valentyna (approved):

D1nviz (approved):

src/hooks/use-filter-query.tsx
```diff
@@ -24,26 +24,27 @@ export const useFilterQuery = <T extends object>({
...
+  const updateFiltersInQuery: UpdateFiltersInQuery<T> = useCallback(
+    (valuesToUpdate) => {
+      for (const key in valuesToUpdate) {
+        const value = valuesToUpdate[key]!
+
+        if (value) {
+          searchParams.set(String(key), String(value))
+        } else {
+          searchParams.delete(String(key))
+        }
```
byavd, line 42:
```suggestion
      valuesToUpdate.forEach((key) => {...})
```
It's better to use forEach here for better code readability.

byavd (approved):

Tolik170 (approved):

src/hooks/use-filter-query.tsx
mhevyk: Sorry, cannot use `foreach` here because `valuesToUpdate` in an object. I think it will be cleaner and more performant because if we use `Object.keys()`, it first compute keys and then we should iterate over them again. So we iterate over keys two times

PR #1781, opened 2024-04-10
Improve coverage for the QuestionEditor component

iva-stasia:
Added tests for the `QuestionEditor` component to improve coverage.

In `QuestionEditor.tsx`:
- had to add `data-testid='addNewAnswerBtn'` to test the function, since there's no other way to get this element
- remove the unnecessary nullish operator in the `setTypeValue` function, since we don't expect any other type of `value` exept `string`

Screenshot 2024-04-10 at 11 58 59

### DIFF
--- a/tests/unit/components/question-editor/QuestionEditor.spec.jsx
+++ b/tests/unit/components/question-editor/QuestionEditor.spec.jsx
@@ -1,33 +1,43 @@
 import QuestionEditor from '~/components/question-editor/QuestionEditor'
 import { screen, fireEvent } from '@testing-library/react'
 import { renderWithProviders } from '~tests/test-utils'
-import { beforeEach, describe } from 'vitest'
+import { afterEach, beforeEach, describe, expect } from 'vitest'
 
-const data = {
-  type: 'openAnswer',
-  title: 'question 1',
-  category: null,
-  text: '',
-  openAnswer: '',
-  answers: []
-}
 const handleInputChange = vi.fn()
 const handleNonInputValueChange = vi.fn()
 const onEditMock = vi.fn()
+const onSave = vi.fn()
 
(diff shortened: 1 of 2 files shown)


OliaHrytsak (approved):

tests/unit/components/question-editor/QuestionEditor.spec.jsx
```diff
@@ -1,33 +1,43 @@
...
-
-const data = {
-  type: 'openAnswer',
-  title: 'question 1',
-  category: null,
-  text: '',
-  openAnswer: '',
-  answers: []
-}
+import { afterEach, beforeEach, describe, expect } from 'vitest'
```
mhevyk, line 4: Please make sure all imports are required

tests/unit/components/question-editor/QuestionEditor.spec.jsx
```diff
@@ -1,33 +1,43 @@
...
+    category: null,
+    text: 'question text',
+    openAnswer: 'answer',
+    answers: []
+  }
+
+  const openQuestionProps = {
+    ...props,
+    data: openQuestionData
+  }
```
mhevyk, line 33: Maybe it would be better to move these constants outside of `describe` block? Because `props` are outside it, but `openQuestionData` and `openQuestionProps` are not

mhevyk: Okay, got it. I didn't notice, that those constants are used only within one describe block. If so, it is good to go.

mhevyk (approved):

D1nviz (approved):

maarekua (approved):

byavd (approved):

PR #1785, opened 2024-04-11
Fixed course grid

GeorgeShvab:
- I used Mui's Grid to fix the course grid. Thanks to Mui Grid, course grid works normally on different screens. To make `CourseCard` responsive in the grid, I added 100% width for it inside of `MyCoursesCardsList`.
- Also, I setted `CourseCard` width to `480px` by default, so it covers one-third of PageWrapper max-width (considering two gaps), as it is in Figma design.

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/62070431/7c99dacc-209e-47eb-9409-a791eafeebc1


src/containers/my-courses/my-courses-container/MyCorsesCardsList.tsx
```diff
@@ -29,18 +29,30 @@ const MyCorsesCardsList: FC<CourseListProps> = ({
   ...props
 }) => {
   const courseItems = items.map((course) => (
-    <Box key={course._id} onClick={() => onCourseSelect?.(course)}>
+    <Grid
```
Tolik170, line 32: Maybe better is to use the css grid and for example if we decide to use another lib, not mui, it should work anyway

GeorgeShvab: I agree. I changed this to a simple css grid.

src/components/course-card/CourseCard.styles.ts
```diff
@@ -24,8 +24,9 @@ export const styles = {
     display: 'flex',
     flexDirection: 'column',
     justifyContent: 'space-between',
-    width: '390px',
-    minHeight: '245px',
+    width: '480px',
+    boxSizing: 'border-box',
+    minHeight: '285px',
```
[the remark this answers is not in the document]
GeorgeShvab: Thanks for suggestion. I removed width: `480px`. Instead, I added min-width `360px` as in figma design. I don't think there would be use-cases with smaller widths than `360px` (it looks weird considering its height).

iva-stasia (approved):

D1nviz (approved):

src/containers/my-courses/my-courses-container/MyCorsesCardsList.styles.ts
```diff
@@ -1,7 +1,11 @@
 export const styles = {
   wrapper: {
-    display: 'flex',
-    flexWrap: 'wrap',
+    display: 'grid',
+    gridTemplateColumns: {
+      xs: '1fr',
+      md: 'repeat(2, 1fr)',
+      lg: 'repeat(3, 1fr)'
+    },
```
mhevyk, line 8: Please consider using handy grid-building auto-sizing functions like following:
```css
grid-template-columns: repeat(auto-fill | auto-fit, minmax(min(200px, 100%), 1fr));
```

I hope it can set min sizes of card and number of cards per view using single css line

Maybe would be useful: https://css-tricks.com/auto-sizing-columns-css-grid-auto-fill-vs-auto-fit/

asya982 (approved):

src/containers/my-courses/my-courses-container/MyCorsesCardsList.styles.ts
GeorgeShvab: I tried to use one-line approach, including provided `grid-template-columns: repeat(auto-fit, minmax(min(200px, 100%), 1fr));`, but I couldn't find the right thing to make it work beautifully and without bugs on different screens and with different amount of courses.

maarekua (approved):

src/components/course-card/CourseCard.styles.ts
```diff
@@ -24,8 +24,9 @@ export const styles = {
     display: 'flex',
     flexDirection: 'column',
     justifyContent: 'space-between',
-    width: '390px',
-    minHeight: '245px',
+    minHeight: '285px',
+    minWidth: '360px',
```
dmtrth25, line 28: If you set `minWidth` for card. It's not looks perfect if we pass too long title for example or description as well. It'll break

image

dmtrth25 (changes requested):

src/components/course-card/CourseCard.styles.ts
GeorgeShvab: I slightly changed grid so it doesn't break when there are very long titles or descriptions. Also, added `break-word` to title and description, so text can go to a new line. There is also added `height: 100%` for cards inside of this grid, so in case one card has bigger height (as result of double line title and description), other cards in the row have the same height too.
image

mhevyk (approved):

src/components/course-card/CourseCard.styles.ts
dmtrth25: > I slightly changed grid so it doesn't break when there are very long titles or descriptions. Also, added `break-word` to title and description, so text can go to a new line. There is also added `height: 100%` for cards inside of this grid, so in case one card has bigger height (as result of double line title and description), other cards in the row have the same height too. image

I think that you can try to write `wordBreak` property only once for your `card` style. 
And not repeat it for `title` and `description` and also can write `height` property and not to create another one

src/containers/my-courses/my-courses-container/MyCorsesCardsList.styles.ts
```diff
@@ -2,11 +2,14 @@ export const styles = {
...
-      md: 'repeat(2, 1fr)',
-      lg: 'repeat(3, 1fr)'
+      xs: 'minmax(0, 1fr)',
+      md: 'repeat(2, minmax(0, 1fr))',
+      lg: 'repeat(3, minmax(0, 1fr))'
     },
     gap: '24px',
     mb: '24px'
+  },
+  card: {
```
dmtrth25, line 12: Try to set `height` for `card` styles as I write above and not to create a new property

src/components/course-card/CourseCard.styles.ts
GeorgeShvab: Thanks for suggestion. I moved `wordBreak` to general card style in recent push. Also moved `height: 100%` directly to card styles, instead of setting it in `MyCoursesCardsList`.

dmtrth25 (approved):

PR #1786, opened 2024-04-11
Increased the coverage of the SubjectsStep component

t-valentyna:


OliaHrytsak (approved):

tests/unit/containers/tutor-home-page/subjects-step/SubjectsStep.spec.jsx
```diff
@@ -0,0 +1,155 @@
...
+  afterEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('should render SubjectsStep title', () => {
+    const title = screen.getByText('becomeTutor.categories.title')
+    expect(title).toBeInTheDocument()
+  })
+
+  it('should add a new subject and delete a subject', async () => {
```
iva-stasia, line 85: As I can see, in this test you are checking not only the deletion and addition of the subject chip, but also the resetting of the subject field value after changing the category.  Maybe it's better to divide this test? Because now it looks a bit overloaded.

tests/unit/containers/tutor-home-page/subjects-step/SubjectsStep.spec.jsx
```diff
@@ -0,0 +1,155 @@
...
+    const addSubject = screen.getByTestId('add-subject')
+    expect(addSubject).toBeInTheDocument()
+
+    const firstField = screen.getByLabelText(
```
iva-stasia, line 89: Variable names might be something like `categoryField' instead of `firstField' to make the code a bit easier to read. The same thing works for the subject field.

t-valentyna: I agree, the variables' names were changed to `categoryField` and `subjectField`.

tests/unit/containers/tutor-home-page/subjects-step/SubjectsStep.spec.jsx
iva-stasia: Following this logic, we can put almost all checks into one test 😅. For example, you can group tests with repeated setup into a separate `describe` block.
But it works, so the rest is up to you.

iva-stasia (approved):

mhevyk (approved):

asya982 (approved):

GeorgeShvab (approved):

maarekua (approved):

PR #1787, opened 2024-04-12
Fix behavior of "Create new question" for quiz button 

OliaHrytsak:
- Daleted unnecessary useEffect
- Added customCloseModal to modal-context.tsx for changing previous behavior of close modal function 

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/127191734/27b751eb-ff45-434f-b37f-242475a09d55

### DIFF
--- a/src/context/modal-context.tsx
+++ b/src/context/modal-context.tsx
@@ -37,9 +37,6 @@ const ModalProvider: FC<ModalProviderProps> = ({ children }) => {
   const [timer, setTimer] = useState<NodeJS.Timeout | null>(null)
   const customCloseModalRef = useRef<(() => void) | undefined>()
 
-  console.log(customCloseModalRef.current)
-  console.log('ModalProvider')
-
   const closeModal = useCallback(() => {
     setModal(null)
     setPaperProps({})
--- a/src/context/modal-context.tsx
+++ b/src/context/modal-context.tsx
@@ -70,7 +70,11 @@ const ModalProvider: FC<ModalProviderProps> = ({ children }) => {
   )
 
   const handleCloseModal = () => {
-    customCloseModal ? customCloseModal() : closeModal()
+    if (customCloseModal) {
+      customCloseModal()
+      closeModal()
+    }
+    closeModal()
   }
 
   return (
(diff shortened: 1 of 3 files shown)


iva-stasia (approved):

D1nviz (approved):

src/context/modal-context.tsx
```diff
@@ -32,6 +34,7 @@ const ModalProvider: FC<ModalProviderProps> = ({ children }) => {
   const [modal, setModal] = useState<ReactElement | null>(null)
   const [paperProps, setPaperProps] = useState<PaperProps>({})
   const [timer, setTimer] = useState<NodeJS.Timeout | null>(null)
+  const [customCloseModal, setCustomCloseModal] = useState<() => void>()
```
mhevyk, line 37: Please consider using `useRef` to store `customCloseModal` instead of `useState` to avoid unnecessary rerenders in context.

src/context/modal-context.tsx
```diff
@@ -67,7 +74,14 @@ const ModalProvider: FC<ModalProviderProps> = ({ children }) => {
       {children}
       {modal && (
         <PopupDialog
-          closeModal={closeModal}
+          closeModal={
```
dmtrth25, line 77: Can create it more readability. 
Implement separate function and just pass it as a prop
`closeModal={handleCloseHandler}`
```ts
const handleCloseHandler = () => {
  customCloseModal ? customCloseModal() : null
  closeModal()
}
```

src/context/modal-context.tsx
OliaHrytsak: customCloseModal is stored inside of modal-context and is a part of close modal behavior, when user click on close icon
When I've changed useState into useRef (for testing) context still rerenders each time, when we click on close icon.

Now, handleCloseModal function looks like that

image

mhevyk: I meant that you have an extra rerender on init phase in following code:

```jsx
paperProps && setPaperProps(paperProps)
delayToClose && closeModalAfterDelay(delayToClose)
customCloseModal && setCustomCloseModal(() => customCloseModal)
```

setState always causes rerenders, so `setCustomCloseModal` will trigger a rerender. If component still rerenders, I think using `useRef` is a better option

mhevyk (approved):

dmtrth25 (approved):

byavd (approved):

Issue #1771, opened 2024-04-08
(SP: 2) Add mixins to design system


asya982:
[Link to course fragment about mixins](https://softserveacademy.udemy.com/course/react-for-senior-engineers/learn/lecture/26837968#questions)

You should think of reusable piecies of "css-functions" that can be used throughout the project, f.e default flex setup, `@media` rules for different resolutions, etc.

##### If you should add some variables - place them inside `_variables.scss` !!!

PR #1788, opened 2024-04-12
Added breakpoints and spacings to design system variables

mhevyk:
- [x] Added variables for breakpoints to file `_variables.scss`.

### Structure of variables:
I organized the structure of variables as following (`breakpoints` example):

```scss
$mobile: 25.875rem; // 414px
$tablet: 55rem; // 880px
$desktop: 90rem; // 1440px

$breakpoints: (
  mobile: $mobile,
  tablet: $tablet,
  desktop: $desktop
);
```

I created a `map` for every variable set for futher usage inside scss `mixins` or `loops` to generate classes.

### Example of possible usage (after defining `mixins` in #1771):

`.scss` file:

```scss
@mixin tablet {
  @media (min-width: map-get($breakpoints, 'tablet')) {
    @content;
  }
}

@mixin desktop {
  @media (min-width: map-get($breakpoints, 'desktop')) {
    @content;
  }
}

@mixin padding($space) {
  padding: map-get($spacing, $space);
}

.test {
  color: black;
  @include padding('md');

  @include tablet {
    color: blue;
  }
  @include desktop {
    color: red;
  }
}
```

Compiled `.css` file:

```scss
.test {
  color: black;
  padding: 1.5rem;
}
@media (min-width: 55rem) {
  .test {
    color: blue;
  }
}
@media (min-width: 90rem) {
  .test {
    color: red;
  }
}
```

### DIFF
--- a/src/scss/src/foundation/_variables.scss
+++ b/src/scss/src/foundation/_variables.scss
@@ -258,6 +258,42 @@ $tt: (
 =                BREAKPOINTS                  =
 =============================================*/
 
-/* Breakpoints */
+$mobile: 25.875rem; // 414px
+$tablet: 55rem; // 880px
+$desktop: 90rem; // 1440px
+
+$breakpoints: (
+  mobile: $mobile,
+  tablet: $tablet,
+  desktop: $desktop
+);
+
+/*=============================================
+=                  SPACING                   =
+=============================================*/
+
+$spacing-none: 0;
+$spacing-xxxs: 0.25rem; // 4px
+$spacing-xxs: 0.5rem; // 8px
+$spacing-xs: 0.75rem; // 12px
+$spacing-sm: 1rem; // 16px
+$spacing-md: 1.5rem; // 24px
+$spacing-lg: 2rem; // 32px
+$spacing-xl: 3rem; // 48px
+$spacing-2xl: 4.5rem; // 72px
+$spacing-3xl: 6rem; // 96px
+
+$spacing: (
+  none: $spacing-none,
+  xxxs: $spacing-xxxs,
+  xxs: $spacing-xxs,
+  xs: $spacing-xs,
+  sm: $spacing-sm,
+  md: $spacing-md,
+  lg: $spacing-lg,
+  xl: $spacing-xl,
+  2xl: $spacing-2xl,
+  3xl: $spacing-3xl
+);
 
 /*=====  End of BREAKPOINTS  ======*/


src/scss/src/foundation/_variables.scss
```diff
@@ -258,6 +258,42 @@ $tt: (
...
+  desktop: $desktop
+);
+
+/*=============================================
+=                  SPACING                   =
+=============================================*/
+
+$spacing-none: 0;
+$spacing-xxxs: 0.25rem; // 4px
+$spacing-xxs: 0.5rem; // 8px
```
iva-stasia, line 277: Don't these spacing keys like `xxxs`, `xxs` and so on look a bit unusual in comparison to other design systems?
Wouldn't it be more convenient to use for spacing something like `spacing-1`, `spacing-2` similar to the [Carbon design system](https://carbondesignsystem.com/elements/spacing/overview/#spacing-scale), for example?

mhevyk: I think T-shirt-like approach will be more suitable because I hope we will create `Margin` and `Padding` or `Spacing` components like following (code from senior react course: https://softserveacademy.udemy.com/course/react-for-senior-engineers/learn/lecture/26838048#questions)

Example of `Margin` component: 
```jsx
import React from 'react'
import { Spacing } from '@s2s/foundation'

export interface MarginProps {
    space?: keyof typeof Spacing
    left?: boolean,
    right?: boolean,
    top?: boolean,
    bottom?: boolean
}

const Margin: React.FC<MarginProps> = ({ space = 'xxxs', children, left, right, top, bottom }) => {
    let className = ``

    if (!left && !right && !top && !bottom) {
        className = `s2s-margin-${space}`
    }

    if (left) {
        className = `${className} s2s-margin-left-${space}`
    }

    if (right) {
        className = `${className} s2s-margin-right-${space}`
    }

    if (top) {
        className = `${className} s2s-margin-top-${space}`
    }

    if (bottom) {
        className = `${className} s2s-margin-bottom-${space}`
    }

    return <div className={className}>
        {children}
    </div>
}

export default Margin
```

And example of usage:

```jsx
<Margin left space="lg">
     <div>Content with large left margin</div>
</Margin>

<Margin right space="xs">
      <div>Content with extra small right margin</div>
</Margin>
```

As you can see we will pass `space` prop, which will be autocompleting sizes of spacing like `xxxs`, `xxs`, and so on. In my opinion `spacing-1`, `spacing-2` options and so on is not very readable when using as a prop

iva-stasia: From this point of view, it makes sense :)

iva-stasia (approved):

asya982 (approved):

maarekua (approved):

D1nviz (approved):

PR #1790, opened 2024-04-12
Feature/1761/edit resourse cooperation

asya982:
### Enhancements:
- Added helper function to extract file extension `getAttachmentExtension`
- Fixed type error in `ResourceItem` for `Box` element
- Added type view in`EditAttachmentModal`

### Implementation:
- Added `editResource` function to `CourseSectionContainer` which is responsible for opening editing form for resources
- Added logic for editing attachemnt on the same page


maarekua (approved):

src/containers/my-resources/edit-attachment-modal/EditAttachmentModal.tsx
iva-stasia, line 1: When trying to change an attachment, I got a validation error that looks strange:
image
It's strange because I successfully uploaded a file named 'cat.jpeg', and now I get an error when I just touch this field.
Since this was implemented before you, it might be a good idea to discuss it in a meeting. Maybe there were some reasons for it. Or you can just change it in your PR. It's up to you, but we must not forget about it :)

asya982: Yeha, i've noticed it too, this modal is a little bit strange in general, because of this manipulation with extension, i think i will fix it here in this pr, since, i've already made some changes there

asya982: There is a room for improvment, but for now i fixed it

iva-stasia (approved):

src/containers/course-section/CourseSectionContainer.tsx
```diff
@@ -188,6 +196,56 @@ const CourseSectionContainer: FC<SectionProps> = ({
...
+
+      window
+        .open(
+          createUrlPath(
+            authRoutes.myResources[navigationFiled].path,
+            resource._id
+          ),
+          '_blank'
+        )
+        ?.focus()
```
mhevyk, line 245: What is `focus` used for here?

asya982: In order to focus user on new tab (edit quiz or lessons)

src/containers/my-resources/edit-attachment-modal/EditAttachmentModal.constants.ts
```diff
@@ -1,25 +1,27 @@
...
 
-export const getInitialValues = (attachment: Attachment) => {
-  const initialFileName = attachment.fileName.substring(
-    0,
-    attachment.fileName.lastIndexOf('.')
-  )
+export const getInitialValues = (
+  attachment: Attachment
+): EditAttachmentForm => {
+  const [initialFileName, fileExtension] = attachment.fileName.split('.')
```
mhevyk, line 7: I think it is not correct solution. For example attachment's filename can be `my.attachment-present-simple.pdf`. If it is not checked before this function is used, variables will have following values:
- `initialFileName`: `my`
- `fileExtension`: `attachment-present-simple.pdf`

Please make sure it is checked before this splitting is used or write better solution to split file to filename and extension

src/utils/helper-functions.tsx
```diff
@@ -288,3 +288,6 @@ export const getGroupedByDate = <T extends { createdAt: string }>(
 
     return result
   }, [])
+
+export const getAttachmentExtension = (fileName: string): string =>
+  fileName.split('.').reverse()[0]
```
mhevyk, line 293: I suggest rewriting this function to split filename into `basename` and `extension`. For instance filename `test.file-bla-bla-bla.pdf` has to be splitted into basename `test.file-bla-bla-bla` and extension `pdf`. Then we can reuse it in `EditAttachmentModal.constants.ts`

mhevyk (changes requested):

t-valentyna (approved):

src/utils/helper-functions.tsx
asya982: rewrote this function according to your notes, thank you!

src/containers/my-resources/edit-attachment-modal/EditAttachmentModal.constants.ts
asya982: according to current fixes, now we have an object returned from this function with `fileName` and `fileExtension`

src/containers/my-resources/edit-attachment-modal/EditAttachmentModal.constants.ts
```diff
@@ -1,25 +1,28 @@
-import { Attachment } from '~/types'
+import { Attachment, EditAttachmentForm } from '~/types'
 import { emptyField, textField } from '~/utils/validations/common'
+import { getAttachmentInfo } from '../../../utils/helper-functions'
```
mhevyk, line 3: use import alias here

src/utils/helper-functions.tsx
```diff
@@ -288,3 +288,13 @@ export const getGroupedByDate = <T extends { createdAt: string }>(
 
     return result
   }, [])
+
+export const getAttachmentInfo = (
```
mhevyk, line 292: I suggest to rename this function to someting like `parseFilename` to help our teammates in finding this method and reuse in other places

mhevyk (changes requested):

src/containers/my-resources/edit-attachment-modal/EditAttachmentModal.constants.ts
mhevyk: Nice work!

src/utils/helper-functions.tsx
asya982: solved

src/containers/my-resources/edit-attachment-modal/EditAttachmentModal.constants.ts
asya982: solved

src/utils/helper-functions.tsx
mhevyk: Became much better!

mhevyk (approved):

PR #1791, opened 2024-04-14
Changed video presentation behaviour

GeorgeShvab:
- Added conditions, so if profile's owner is tutor, VideoPresentation will always be shown. If owner is student, than VideoPresentation will be shown only if he has videoLink.
- Added test cases.

Behaviour in video, firstly for student without video link then after link was added. Also for tutor without link. 

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/62070431/12acf3f0-7e02-4dc5-a548-5e062c746e6d

### DIFF
--- a/src/pages/tutor-profile/TutorProfile.jsx
+++ b/src/pages/tutor-profile/TutorProfile.jsx
@@ -16,6 +16,8 @@ import { defaultResponses } from '~/constants'
 import { responseMock } from '~/pages/tutor-profile/constants'
 import AboutTutorBlock from '~/containers/tutor-profile/about-tutor-block/AboutTutorBlock'
 
+import { UserRoleEnum } from '~/types'
+
 const TutorProfile = () => {
   const { id } = useParams()
   const [searchParams] = useSearchParams()
@@ -25,9 +27,12 @@ const TutorProfile = () => {
 
   const { userId, userRole } = useAppSelector((state) => state.appMain)
 
+  const preferredRole = paramsRole || userRole
+  const preferredId = id || userId
+
   const getUserData = useCallback(
-    () => userService.getUserById(id || userId, paramsRole || userRole),
-    [userId, userRole, id, paramsRole]
+    () => userService.getUserById(preferredId, preferredRole),
+    [preferredId, preferredRole]
   )
 
   const { loading, response } = useAxios({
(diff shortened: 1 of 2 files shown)


tests/unit/pages/tutor-profile/TutorProfile.spec.jsx
```diff
@@ -76,12 +84,60 @@ describe('TutorProfile', () => {
...
+    )
+
+    expect(videoBlockTitle).not.toBeInTheDocument()
+  })
+
+  it('Should render video presentation block when student has a link', () => {
+    const fakeData = getFakeData(false, videoMockData)
+
+    useAxios.mockImplementation(() => fakeData)
+    renderWithProviders(<TutorProfile />, {
```
iva-stasia, line 132: As I noticed, in each test we need to do the same setup, but with different parameters. I guess you can move this common logic to a separate function:
```javascript
const renderWithMockData = (load, initialEntries, extraData = {}) => {
  //mock and rendering logic
}
```
And then just reuse it:
```javascript
it('Should render video presentation block when student has a link', () => {
  renderWithMockData(false, studentRoute, videoMockData)
  ...
})
```

tests/unit/pages/tutor-profile/TutorProfile.spec.jsx
```diff
@@ -76,12 +84,60 @@ describe('TutorProfile', () => {
...
+    expect(videoBlockTitle).toBeInTheDocument()
+  })
+
+  it('Should not render video presentation block when student has no video link', () => {
+    const fakeData = getFakeData(false)
+
+    useAxios.mockImplementation(() => fakeData)
+    renderWithProviders(<TutorProfile />, {
+      initialEntries: studentRoute,
+      preloadedState: { appMain }
```
OliaHrytsak, line 118: In this test you test behavior, when role is student, but use the same preloadedState (tutor)

src/pages/tutor-profile/TutorProfile.jsx
```diff
@@ -45,7 +52,9 @@ const TutorProfile = () => {
       <ProfileInfo myRole={userRole} userData={response} />
       <CompleteProfileBlock data={response} profileItems={profileItems} />
       <AboutTutorBlock />
-      <VideoPresentation />
+      {(isTutor || (!isTutor && response.videoLink?.student)) && (
```
mhevyk, line 55: Please consider creating variable `shouldShowPresentation` and move this logic up to avoid messing up our `jsx` code

tests/unit/pages/tutor-profile/TutorProfile.spec.jsx
GeorgeShvab: User role for tests was taken from `studentRoute` and `tutorRoute` variables. It worked, as inside of `TutorProfile` there is such condition: `paramsRole || userRole` (paramsRole is from search params, userRole is from redux appMain). But it really can be confusing, so in tests, I removed roles from routes and instead set them in appMain.

tests/unit/pages/tutor-profile/TutorProfile.spec.jsx
GeorgeShvab: Thanks. I changed it as you suggested in a new commit.

src/pages/tutor-profile/TutorProfile.jsx
GeorgeShvab: Thanks for the suggestion, I moved it.

asya982 (approved):

mhevyk (approved):

iva-stasia (approved):

PR #1792, opened 2024-04-14
Add coverage for cooperation notes and cooperation activities list

iva-stasia:
1. Added tests for the `CooperationActivitiesList` component:
Screenshot 2024-04-13 at 00 59 09

2. Added tests for the `CooperationNotes` component:
Screenshot 2024-04-13 at 21 26 04

The uncovered branches:
- This branch looks illogical, because reaching this branch means that an empty snackbar will be displayed. A new issue will be created to investigate and fix the bug.
Screenshot 2024-04-13 at 21 25 47

- This branch indicates a type issue. According to the `deleteNote` function in `CooperationNotesService` the parameters are required, but `ServiceFunction` type that we need to pass to the `useAxios` hook requires a union type like `type | undefined`. As I understand it, this solution was created to solve this issue. But passing an empty string is't a good idea, since the `deleteNote` function expects real ids. Therefore, a new issue will be created to investigate and fix the type issue. The `ServiceFunction` function type is already changed in #1795.
Screenshot 2024-04-13 at 19 11 54

### DIFF
--- a/tests/unit/containers/my-cooperations/cooperation-activities-list/CooperationActivitiesList.spec.jsx
+++ b/tests/unit/containers/my-cooperations/cooperation-activities-list/CooperationActivitiesList.spec.jsx
@@ -1,43 +1,226 @@
-import { screen, fireEvent } from '@testing-library/react'
+import { screen, fireEvent, waitFor } from '@testing-library/react'
 import CooperationActivitiesList from '~/containers/my-cooperations/cooperation-activities-list/CooperationActivitiesList'
 import { renderWithProviders } from '~tests/test-utils'
+import { sectionInitialData } from '~/pages/create-course/CreateCourse.constants'
 
-const mockedData = {
+const originalDateNow = Date.now
+Date.now = () => 1487076708000
+
+const { useCooperationContextMock } = vi.hoisted(() => {
+  return {
+    useCooperationContextMock: vi.fn()
+  }
+})
+
+vi.mock('~/context/cooperation-context', async () => {
+  const actual = await vi.importActual('~/context/cooperation-context')
+  return {
+    ...actual,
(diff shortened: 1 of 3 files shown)


tests/unit/containers/my-cooperations/cooperation-activities-list/CooperationActivitiesList.spec.jsx
```diff
@@ -1,43 +1,224 @@
...
+      expect(titleInput.value).toBe(newTitle)
+    })
+  })
+
+  it('should set sections from the data first in order if the current section index is undefined', async () => {
+    renderWithMockData(mockedSectionsData, undefined)
+
+    const newSectionData = mockedCourseData.sections.map((section, index) => ({
+      ...section,
+      id: Date.now().toString() + index
```
mhevyk, line 144: I have noticed that code piece `Date.now().toString()` is repeated frequently. Do we really need to mock `Date.now()` to then generate id with `Date.now().toString()`? Can't we use just hardcoded `id` instead?

iva-stasia: I can't use hardcoded ids here because `newSectionData` is created in the component using the following function:
```javascript
const newSectionData = selectedCourse.sections.map(
  (section, index) => ({
    ...section,
    id: Date.now().toString() + index
  })
)
```
And I need to check that the data passed to `handleNonInputValueChange` in the test is the same as in the component. To do this, I have to reproduce the function's behavior.

mhevyk (approved):

asya982 (approved):

maarekua (approved):

D1nviz (approved):

Issue #1739, opened 2024-04-04
(SP: 0.5) Missing ukrainian translation of some captions and messages


GeorgeShvab:
Some captions and messages have only english translations. A few ukrainian translations are missing at least in `translations/ua/common.js` and `translations/ua/error-page.js`, compared to corresponding files with english translations.

Issue #1727, opened 2024-04-03
(SP:0.5) Refactor FileUploader to emit action when the same file is added twice in a row


asya982:
When tying to add the same file twice button doesn't make any actions, first time it emits error/request, but if trying to add the same file ance again - nothing happens (function addFiles from useUpload just doesn't revoke)
https://github.com/ita-social-projects/SpaceToStudy-Client/assets/105641724/6cb75204-7288-4576-ac9c-552759814db8

PR #1795, opened 2024-04-14
Add profile general tab functionality

iva-stasia:
Added profile general tab functionality.
- `FileUploader.tsx`: the `accept` attribute was added to prevent adding inappropriate file types. `initialState` and `initialError` parameters were marked as optional, since they have default values and can be optional.
- `services.types.ts`: the type of `ServiceFunction` parameters was changed. In the previous implementation, we were obliged to pass parameters only of the union type like `type | undefined`, which led to many illogical cases (one of them is described in #1792). Now we can pass any type of parameter or not pass any, if needed (according to what this type was designed for).
- `ProfileTab` component was created as a general wrapper for all tabs of the Profile tab (General, Professional, etc.), to wrap all tabs in the `ProfileProvider` of the `profile-contex`.
- `ProfileTabInnerContainer` component was created as a wrapper for all tabs of the Profile tab, where the data saving logic is executed. Styles were added accordingly.
- To the `ProfileGeneralTab` component was added the logic for uploading a photo and saving data.
- `profile-context.tsx` was created to handle the logic for storing data between tabs of the Profile tab. Since the next step in implementing the Profile tab is to add other tabs, I created a context that will be used for all tabs.
- In order not to overload the `ProfileTabInnerContainer`, I moved the logic for updating user data to the `use-update-user` hook.
- The function for getting user initial values was moved from `profile-general-tab/ProfileGeneralTab.constants.ts` to `utils/get-profile-context-initial-values.ts`.
- Validation for YouTube video link was created and added to `utils/validations/common.ts`.
- Translation files were updated with corresponding messages.
- `EditProfile.spec.jsx` was updated according to new data.

To avoid overloading this PR, tests for new files and components will be added in separate PRs.

Result:

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/110097862/401c0bde-ecc5-4310-a650-93a8ec802f99


src/components/file-uploader/FileUploader.tsx
```diff
@@ -70,12 +70,20 @@ const FileUploader: FC<FileUploaderProps> = ({
...
       {buttonText}
       {icon}
-      <input hidden multiple onChange={addFiles} type={InputEnum.File} />
+      <input
+        accept={acceptableFileTypes}
+        hidden
+        multiple
+        onChange={addFiles}
+        type={InputEnum.File}
+      />
```
mhevyk, line 86: If I upload an image, filename of which has more than 55 characters, i receive an error. It is expected behaviour. If I try to upload same file one more time, I don't receive an error as you can see in the video below:

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/109922101/5bdde3ac-972a-46b8-aa71-35af86a0ef21

Please consider fixing it within current PR or creating new issue for it

src/utils/validations/common.ts
```diff
@@ -76,6 +76,16 @@ export const textField =
     }
   }
 
+export const youtubeVideoLink = (value: string): string => {
+  const youtubeRegex =
+    /^(?:https?:\/\/)?(?:m\.|www\.)?(?:youtu\.be\/|youtube\.com\/(?:embed\/|v\/|watch\?v=|shorts\/|watch\?.+&v=))((\w|-){11})(?:\S+)?$/
```
mhevyk, line 81: This pattern is not matching all youtube links. For example, the following link is not considered valid, but in reality it has to be valid:
https://youtu.be/watch?v=rhTl_OyehF8&list=RDrhTl_OyehF8&start_radio=1

image

I used ChatGpt and it helps to test different shapes of youtube link, you can use it too to be more precise in your regex:

image

src/containers/edit-profile/profile-tab/profile-tabs/profile-general-tab/ProfileGeneralTab.tsx
```diff
@@ -181,9 +252,15 @@ const ProfileGeneralTab: FC<ProfileGeneralTabProps> = ({ user }) => {
...
-            )
+            ),
+            sx: styles.videoLinkInput
           }}
+          errorMsg={t(errors.videoLink)}
           fullWidth
+          onBlur={handleBlur('videoLink')}
+          onChange={handleInputChange('videoLink')}
+          placeholder='youtube.com/my-video'
+          value={data.videoLink?.replace('https://', '')}
```
mhevyk, line 263: I think we should provide default value if our backend will not send us data.videoLink to increase safety. Example:
```suggestion
          value={data.videoLink?.replace('https://', '') ?? ""}
```

src/containers/edit-profile/profile-tab/profile-tabs/profile-general-tab/ProfileGeneralTab.tsx
```diff
@@ -38,47 +55,101 @@ const ProfileGeneralTab: FC<ProfileGeneralTabProps> = ({ user }) => {
...
+      ...updatedData,
+      photo: null
+    })
+  }
 
+  const photoToDisplay =
+    photo === null
+      ? ''
+      : photo?.src ||
+        (user.photo && `${import.meta.env.VITE_APP_IMG_USER_URL}${user.photo}`)
```
mhevyk, line 129: Maybe use `createUrlPath` helper function here instead?

mhevyk (changes requested):

src/components/file-uploader/FileUploader.tsx
iva-stasia: The issue for this bug has already been created: #1727.
That's why I think it is better to fix this bug in that issue.

mhevyk: Okay, I did not remember about it. Thanks!

src/utils/validations/common.ts
iva-stasia: ✅ 
I simplified the pattern so that it only checks for the YT domain name. Now it accepts a wider range of YT links.

src/containers/edit-profile/profile-tab/profile-tabs/profile-general-tab/ProfileGeneralTab.tsx
iva-stasia: After examining this line, I realized that the optional chaining operator is redundant here, since I'm already checking the value in [line 48](https://github.com/ita-social-projects/SpaceToStudy-Client/pull/1795/files#diff-a6ecf36f69ddeb793427e9d657db2d5934f6fe82a40bf5381432c8f87a63cc60R48).
So I simply removed the optional chaining operator.

src/containers/edit-profile/profile-tab/profile-tabs/profile-general-tab/ProfileGeneralTab.tsx
iva-stasia: Unfortunately, it doesn't suit here. 
Current result looks something like: `path/photo`. Using the function, I get an extra slash: `path//photo`.
It's because the `VITE_APP_IMG_USER_URL` variable contains a slash at the end. But I believe it's better not to change the variable or at least to do it in a separate PR, as it requires changes in many files.

mhevyk: Our helper function is a bit strange. I would refactor it to concat urls correctly, for example using `new URL(path, base).href`, which concats urls correctly regardless of slash at the end. Possible candidate for a new issue? :)

src/utils/validations/common.ts
mhevyk: The regular expression became less complicated, which is great!

mhevyk (approved):

Mygal-Alex (approved):

src/constants/translations/ua/common.json
```diff
@@ -51,6 +51,9 @@
     "moveToDraft": "Перемістити до чернетки",
     "closeOffer": "Закрити пропозицію"
   },
+  "errorMessages": {
+    "youtubeLink": "Будь ласка, надайте дійсне посилання на YouTube"
+  },
```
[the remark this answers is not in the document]
iva-stasia: You are right. And now I noticed that this is true for other translations as well (for instance, labels). And some messages should be changed. For example, `This text is too short` isn't a good example of an error message.
So maybe I should create a new issue to check and correct all inconsistencies? What do you think?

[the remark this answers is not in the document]
asya982: Yeah, we a;ready have it - #1739, so I think you can just add those, which was added by you in this pr

asya982 (approved):

PR #1798, opened 2024-04-15
Feature/1771/scss mixins

asya982:
Added a collection of mixins to enhance the maintainability and flexibility of styling across components. There are mixins for handling responsive design using media queries (tablet, desktop, mobile), spacing utilities (spacing mixin), container sizing shortcuts (box mixin), flexbox and grid layout helpers (flexbox and grid mixins), text manipulation tools (text-truncate mixin), and useful resets (listReset and clearFix mixins).

#### Usage Examples:
```
@include tablet { ... } - Applies styles within a tablet media query.
@include spacing(margin, 'lg', 'xxs') - Sets up margin vertically for 32px and horizontally for 8px.
@include box(1rem, 2rem) - Defines width as 1rem and height as 2rem.
@include flexbox(column, space-between) - Sets up a flexbox container with column direction and space-between alignment.
@include text-truncate('...') - Truncates text with a custom ellipsis string.
@include listReset - Resets margin, padding, and list-style for <ul> elements.
@include clearFix - Creates a clearfix for containing floated elements properly.
```

### DIFF
--- a/src/scss/src/foundation/_mixins.scss
+++ b/src/scss/src/foundation/_mixins.scss
@@ -1,4 +1,3 @@
-@use "sass:list";
 /*=============================================
 =                 #MIXINS                     =
 =============================================*/
@@ -130,8 +129,11 @@
  */
 
 @mixin text-truncate($line-clamp: 1) {
+    @if $line-clamp != round($line-clamp) {
+        @error "#{$line-clamp} is not a valid property. Expected an plain integer.";
+    }
     display: -webkit-box;
-    -webkit-line-clamp: $line-clamp; // values are: integer or none;
+    -webkit-line-clamp: $line-clamp; // value is: integer;
     -webkit-box-orient: vertical;
 }
 
@@ -153,26 +155,29 @@
  * 
  */
 
-@mixin font-setup-rubik($size: false, $color: false, $weight: false, $height: false, $letter-space: false) {
+@mixin font-setup-rubik($font-size: false, $color: false, $font-weight: false, $line-height: false, $letter-spacing: false) {
     font-family: $rubik;
-    @if $size {
-        $extractedSize: map-get($fs, $size);
+    @if $font-size {
+        $allSizes: map-merge($fs, $fs-h);
+        $extractedSize: map-get($allSizes, $font-size);
         font-size: $extractedSize;
     }
     @if $color {
         color: $color;
     }
-    @if $weight {
-        $extractedWeight: map-get($fw, $weight);
+    @if $font-weight {
+        $extractedWeight: map-get($fw, $font-weight);
         font-weight: $extractedWeight;
     }
-    @if $height {
-        $extractedHeight: map-get($lh, $height);
+    @if $line-height {
+        $allHeights: map-merge($lh, $lh-h);
+        $extractedHeight: map-get($allHeights, $line-height);
         line-height: $extractedHeight;
     }
-    @if $letter-space {
-        $extractedSpace: map-get($ls, $letter-space);
-        letter-spacing: $extractedSpace;
+    @if $letter-spacing {
+        $allSpacings: map-merge($ls, $ls-h);
+        $extractedSpacing: map-get($allSpacings, $letter-spacing);
+        letter-spacing: $extractedSpacing;
     }
 }
 
--- a/src/scss/src/foundation/_mixins.scss
+++ b/src/scss/src/foundation/_mixins.scss
@@ -1,8 +1,7 @@
+@use "sass:list";
 /*=============================================
 =                 #MIXINS                     =
 =============================================*/
-
-
 
 /**
  *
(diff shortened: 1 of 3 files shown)


src/scss/src/foundation/_mixins.scss
```diff
@@ -1,16 +1,169 @@
...
+ * 
+ * To clip the text without any ellipsis: @include text-truncate('clip');
+ * 
+ */
+
+@mixin text-truncate($overflow: ellipsis) {
+    overflow: hidden;
+    text-overflow: $overflow; // values are: clip, ellipsis, or a string
+    white-space: nowrap;
+}
```
mhevyk, line 134: I suggest adding `text-clamp` mixin to limit number of lines displayed in paragragh

src/scss/src/foundation/_mixins.scss
```diff
@@ -1,16 +1,169 @@
...
+
+    $propertyValues: ();
 
+    @each $value in $values {
+        $spacingValue: map-get($spacing, $value);
+        $propertyValues: append($propertyValues, $spacingValue);
+    }
 
+    #{$property}: $propertyValues;
+}
```
mhevyk, line 60: This method has a bit hard to understand interface. While reading example `@include spacing(margin, 'lg', 'xxs')` I did not get the point what this mixin is actually doing. I suggest splitting this mixin into multiple mixins like following:

```scss
@mixin margin($value);
@mixin margin-push($direction: left, right, top, bottom, $value);
@mixin margin-x($value); // (use margin right and margin left here because we want to create as modular code as possible)
@mixin margin-y($value) // same here;
```

Same for padding

src/scss/src/foundation/_mixins.scss
```diff
@@ -1,16 +1,169 @@
...
+
+/**
+ * mixin is used to reset the margin, padding, and list-style of the <ul> element.
+ *
+ * example usage: @include listReset;
+ *
+ */
+
+@mixin listReset {
+    margin: 0;
```
mhevyk, line 148: Use cebab case naming convention for mixins

src/scss/src/foundation/_mixins.scss
```diff
@@ -1,16 +1,169 @@
...
+@mixin clearFix($content: "") {
+    &:before,
+    &:after {
+        content: $content;
+        display: table;
+    }
+    &:after {
+        clear: both;
+    }
+}
```
mhevyk, line 169: In which scenarios we can use this mixin?

src/scss/src/foundation/_variables.scss
```diff
@@ -297,3 +297,20 @@ $spacing: (
...
+$mainShadow:
+  0px 5px 6px -3px rgba(144, 164, 174, 0.2),
+  0px 9px 12px 1px rgba(144, 164, 174, 0.14),
+  0px 3px 16px 2px rgba(144, 164, 174, 0.12);
+
+$commonShadow: 0px 3px 16px 2px rgba(144, 164, 174, 0.12);
+
+$commonHoverShadow: 0px 3px 16px 2px rgba(144, 164, 174, 0.56);
+
+$smallHoverShadow: 0 3px 1px rgba(144, 164, 174, 0.56);
```
mhevyk, line 314: Also use cebab case for naming shadows variables

src/scss/src/foundation/_mixins.scss
```diff
@@ -1,16 +1,169 @@
...
+
+@mixin flexbox($direction: row, $justify: center, $align: center, $wrap: nowrap, $gap: normal) {
+    display: flex;
+    flex-direction: $direction;
+    justify-content: $justify;
+    align-items: $align;
+    flex-wrap: $wrap;
+    gap: $gap;
+}
+
```
mhevyk, line 98: Does this mixin and others provide default values for unused arguments or properties are added with an error?

src/scss/src/foundation/_mixins.scss
```diff
@@ -1,16 +1,169 @@
...
  *
  */
 
+/*=============================================
+ =                @media rules                =
+ =============================================*/
+
+@mixin tablet {
+    @media (min-width: map-get($breakpoints, "tablet")) {
+        @content;
```
mhevyk, line 19: Maybe you can use similar check for invalid argument like you did below?

```scss
@if not list.index($valid-properties, $property) {
    @error "#{$property} is not a valid property. Expected one of #{$valid-properties}.";
}
```

mhevyk (changes requested):

src/scss/src/foundation/_mixins.scss
asya982: All values which are followed by the semicolons are default values
F.e `$direction: row` - if not provided will be set to `row`
There is an example hwo we can use this mixin, either like a standart js function or manually ype only those arguments which we want to set

src/scss/src/foundation/_mixins.scss
asya982: It's basically help container to adjust it's height and width so all overflowing children would be inside container
[Example](https://www.educative.io/answers/what-is-a-clearfix-in-css)

src/scss/src/foundation/_mixins.scss
asya982: These mixins are not accepting any parameters
you add it like
`@include tablet {
 color: pink;
}`

src/scss/src/foundation/_mixins.scss
asya982: forgot about it, solved

src/scss/src/foundation/_variables.scss
asya982: makes sense, solved

src/scss/src/foundation/_mixins.scss
asya982: I think it's pretty self-explanatory, it's the same as you'd write default margin with 1/2/3/4 parameter, but instead of `rem/px` this mixin accepts our custom breakpoints only

mhevyk: No, i persuade that my approach will be more readable and easier to understand

### Example 1:

For example you want to add `margin` `xxs` on the `left` and `lg` on the `bottom`:

Current implemention:

```scss
@include spacing(margin, 'none', 'none', 'lg', "xxs");
```

Suggested implementation:

```scss
@include margin-push($left, 'xxs');
@include margin-push($bottom, 'lg');
```

### Example 2:

I want to add horizontal `padding` of `lg`:

Current implemention:

```scss
@include spacing(padding, 'none', 'lg');
```

Suggested implemention:

```scss
@include padding-x('lg');
```

src/scss/src/foundation/_mixins.scss
```diff
@@ -1,16 +1,179 @@
...
+}
 
+/**
+ * Setup for grid container
+ *    
+ * @include grid-container(1fr 1fr, 30px, start, center);
+ *
+ */
+
+@mixin grid($columns: auto, $rows: auto, $gap: 20px, $justify: center, $align: center) {
```
mhevyk, line 116: Please use predefined scss variables instead of hardcoding value of 20px

src/scss/src/foundation/_mixins.scss
```diff
@@ -1,16 +1,179 @@
...
+}
+
+/**
+ * mixin is applied to create a clearfix, ensuring that it contains floated elements properly.
+ *
+ * example usage: @include clearFix;
+ *
+ */
 
+@mixin clear-fix($content: "") {
```
mhevyk, line 170: I think `clearfix` should be single word, for instance in article you provided it is single word

mhevyk (changes requested):

mhevyk (approved):

iva-stasia (approved):

t-valentyna (approved):

byavd (approved):

PR #1809, opened 2024-04-17
Implement a video player

Mygal-Alex:
https://github.com/ita-social-projects/SpaceToStudy-Client/assets/101183408/1a726d02-caef-48ca-9212-b7dbc347f8ca

### DIFF
--- a/src/components/video-box/VideoBox.tsx
+++ b/src/components/video-box/VideoBox.tsx
@@ -1,19 +1,25 @@
 import { FC } from 'react'
 import Box from '@mui/material/Box'
-
 import titleBar from '~/assets/img/guest-home-page/title-bar.png'
 import { styles } from '~/components/video-box/VideoBox.styles'
+import VideoPlayer from '../video-player/VideoPlayer'
 
 interface VideoBoxProps {
-  video: string
+  video: string,
+  haveVideo : boolean
 }
 
-const VideoBox: FC<VideoBoxProps> = ({ video }) => {
+const VideoBox: FC<VideoBoxProps> = ({ video, haveVideo }) => {
+  const videoBloc = haveVideo ? 
+  <VideoPlayer video='www.youtube.com/watch?v=LXb3EKWsInQ&t=101s&ab_channel=Jacob%2BKatieSchwarz'/> 
+  :  
+  <Box alt='Video' component='img' src={video} sx={styles.video} /> 
+
   return (
     <Box data-testid='videoBox' sx={styles.videoBox}>
       <Box alt='Title bar' component='img' src={titleBar} />
       <Box sx={styles.videoBg}>
-        <Box alt='Video' component='img' src={video} sx={styles.video} />
+        {videoBloc}
       </Box>
     </Box>
   )
--- a/src/components/video-box/VideoBox.tsx
+++ b/src/components/video-box/VideoBox.tsx
@@ -2,25 +2,25 @@ import { FC } from 'react'
 import Box from '@mui/material/Box'
 import titleBar from '~/assets/img/guest-home-page/title-bar.png'
 import { styles } from '~/components/video-box/VideoBox.styles'
+import videoImg from '~/assets/img/tutor-profile-page/presentationVideoImg.png'
 import VideoPlayer from '../video-player/VideoPlayer'
 
 interface VideoBoxProps {
-  video: string,
-  haveVideo : boolean
+  video: string
+  videoPreview: boolean
 }
 
-const VideoBox: FC<VideoBoxProps> = ({ video, haveVideo }) => {
-  const videoBloc = haveVideo ? 
-  <VideoPlayer video='www.youtube.com/watch?v=LXb3EKWsInQ&t=101s&ab_channel=Jacob%2BKatieSchwarz'/> 
-  :  
-  <Box alt='Video' component='img' src={video} sx={styles.video} /> 
+const VideoBox: FC<VideoBoxProps> = ({ video, videoPreview }) => {
+  const videoBloc = videoPreview ? (
+    <Box alt='Video' component='img' src={videoImg} sx={styles.video} />
+  ) : (
+    <VideoPlayer video={video} />
+  )
 
   return (
     <Box data-testid='videoBox' sx={styles.videoBox}>
       <Box alt='Title bar' component='img' src={titleBar} />
-      <Box sx={styles.videoBg}>
-        {videoBloc}
-      </Box>
+      <Box sx={styles.videoBg}>{videoBloc}</Box>
     </Box>
   )
 }
--- a/src/components/video-player/VideoPlayer.styles.ts
+++ b/src/components/video-player/VideoPlayer.styles.ts
@@ -1,11 +1,11 @@
 export const styles = {
-    playerWrapper: {
-        position: 'relative',
-        paddingTop: '56.25%',
-    },
-    player: {
-        position:'absolute',
-        top:'0',
-        left:'0',
-    }
+  playerWrapper: {
+    position: 'relative',
+    paddingTop: '56.25%'
+  },
+  player: {
+    position: 'absolute',
+    top: '0',
+    left: '0'
+  }
 }
--- a/src/components/video-player/VideoPlayer.styles.ts
+++ b/src/components/video-player/VideoPlayer.styles.ts
@@ -1,0 +1,11 @@
+export const styles = {
+    playerWrapper: {
+        position: 'relative',
+        paddingTop: '56.25%',
+    },
+    player: {
+        position:'absolute',
+        top:'0',
+        left:'0',
+    }
+}
--- a/src/containers/tutor-profile/video-presentation/VideoPresentation.jsx
+++ b/src/containers/tutor-profile/video-presentation/VideoPresentation.jsx
@@ -2,9 +2,8 @@ import { useTranslation } from 'react-i18next'
 import Box from '@mui/material/Box'
 import Typography from '@mui/material/Typography'
 import VideoBox from '~/components/video-box/VideoBox'
-import videoImg from '~/assets/img/tutor-profile-page/presentationVideoImg.png'
 
-const VideoPresentation = () => {
+const VideoPresentation = (video, haveVideo) => {
   const { t } = useTranslation()
 
   return (
@@ -12,7 +11,7 @@ const VideoPresentation = () => {
       <Typography sx={{ my: '32px', typography: { md: 'h4', xs: 'h5' } }}>
         {t('tutorProfilePage.videoPresentation.title')}
       </Typography>
-      <VideoBox video={videoImg} />
+      <VideoBox video={video} haveVideo = {haveVideo}/>
     </Box>
   )
 }
--- a/src/containers/tutor-profile/video-presentation/VideoPresentation.jsx
+++ b/src/containers/tutor-profile/video-presentation/VideoPresentation.jsx
@@ -3,15 +3,14 @@ import Box from '@mui/material/Box'
 import Typography from '@mui/material/Typography'
 import VideoBox from '~/components/video-box/VideoBox'
 
-const VideoPresentation = (video, haveVideo) => {
+const VideoPresentation = (props) => {
   const { t } = useTranslation()
-
   return (
     <Box className='section' sx={{ flexDirection: 'column' }}>
       <Typography sx={{ my: '32px', typography: { md: 'h4', xs: 'h5' } }}>
         {t('tutorProfilePage.videoPresentation.title')}
       </Typography>
-      <VideoBox video={video} haveVideo = {haveVideo}/>
+      <VideoBox video={props.video} videoPreview={props.videoPreview} />
     </Box>
   )
 }
--- a/src/pages/tutor-profile/TutorProfile.jsx
+++ b/src/pages/tutor-profile/TutorProfile.jsx
@@ -12,7 +12,7 @@ import { userService } from '~/services/user-service'
 import useAxios from '~/hooks/use-axios'
 
 import { profileItems } from '~/components/profile-item/complete-profile.constants'
-import { defaultResponses } from '~/constants'
+import { defaultResponses, tutor } from '~/constants'
 import { responseMock } from '~/pages/tutor-profile/constants'
 import AboutTutorBlock from '~/containers/tutor-profile/about-tutor-block/AboutTutorBlock'
 
@@ -45,6 +45,7 @@ const TutorProfile = () => {
     if (
       !loading &&
       response &&
+      response.videoLink &&
       response.videoLink.tutor !== null &&
       response.videoLink.tutor !== ''
     ) {
@@ -59,18 +60,24 @@ const TutorProfile = () => {
   const isTutor = preferredRole === UserRoleEnum.Tutor
   const shouldShowPresentation =
     isTutor || (!isTutor && response.videoLink?.student)
-
+  const videoBlock =
+    userRole === tutor ? (
+      <VideoPresentation
+        video={response.videoLink.tutor}
+        videoPreview={videoPreview}
+      />
+    ) : (
+      <VideoPresentation
+        video={response.videoLink.student}
+        videoPreview={videoPreview}
+      />
+    )
   return (
     <PageWrapper>
       <ProfileInfo myRole={userRole} userData={response} />
       <CompleteProfileBlock data={response} profileItems={profileItems} />
       <AboutTutorBlock />
-      {shouldShowPresentation && (
-        <VideoPresentation
-          video={response.videoLink.tutor}
-          videoPreview={videoPreview}
-        />
-      )}
+      {shouldShowPresentation && videoBlock}
       <CommentsWithRatingBlock
         averageRating={response?.averageRating?.tutor}
         reviewsCount={reviews}
--- a/src/pages/tutor-profile/TutorProfile.jsx
+++ b/src/pages/tutor-profile/TutorProfile.jsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react'
+import { useCallback, useState, useEffect } from 'react'
 import { useAppSelector } from '~/hooks/use-redux'
 import { useParams, useSearchParams } from 'react-router-dom'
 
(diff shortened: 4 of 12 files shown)


src/pages/tutor-profile/TutorProfile.jsx
```diff
@@ -41,20 +41,43 @@ const TutorProfile = () => {
...
   })
 
+  useEffect(() => {
+    if (
+      !loading &&
+      response &&
+      response.videoLink &&
+      response.videoLink.tutor !== null &&
+      response.videoLink.tutor !== ''
+    ) {
```
D1nviz, line 51: This condition looks a bit confusing. Can we refactor to this ?
`!loading && response?.videoLink?.tutor``

Mygal-Alex: I agree. 
we can replace it with this expression 
!loading &&
 response?.videoLink?.tutor !== null &&
 response?.videoLink?.tutor !== ' '
but we can't stop checking for null and empty string.

src/components/video-box/VideoBox.tsx
```diff
@@ -1,20 +1,26 @@
...
+import VideoPlayer from '../video-player/VideoPlayer'
 
 interface VideoBoxProps {
   video: string
+  videoPreview: boolean
 }
 
-const VideoBox: FC<VideoBoxProps> = ({ video }) => {
+const VideoBox: FC<VideoBoxProps> = ({ video, videoPreview }) => {
+  const videoBloc = videoPreview ? (
```
mhevyk, line 14: Please consider using more descriptive variable name instead of `videoBloc`

src/pages/tutor-profile/TutorProfile.jsx
mhevyk: You can also split this expression into separate variables to improve code readability

src/pages/tutor-profile/TutorProfile.jsx
```diff
@@ -41,20 +41,43 @@ const TutorProfile = () => {
...
+      <VideoPresentation
+        video={response.videoLink.tutor}
+        videoPreview={videoPreview}
+      />
+    ) : (
+      <VideoPresentation
+        video={response.videoLink.student}
+        videoPreview={videoPreview}
+      />
+    )
```
mhevyk, line 74: I think these components are pretty similar. I suggest refactoring it as following:

```suggestion   
    <VideoPresentation
        video={response.videoLink[condition here...}
        videoPreview={videoPreview}
      />
```

Also I noticed line of code `const isTutor = preferredRole === UserRoleEnum.Tutor`. Please consider reusing it or other code if possible

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: fixed 
image

src/pages/tutor-profile/TutorProfile.jsx
```diff
@@ -41,6 +41,13 @@ const TutorProfile = () => {
     defaultResponse: defaultResponses.array
   })
 
+  useEffect(() => {
+    const tutorVideo = response?.videoLink?.tutor
+    if (!loading && tutorVideo !== null && tutorVideo !== '') {
```
iva-stasia, line 46: Why can't we simplify it?
```suggestion
    if (!loading && tutorVideo) {
```

tests/unit/components/video-box/VideoBox.spec.jsx
```diff
@@ -1,19 +1,27 @@
...
-
 describe('VideoBox component', () => {
-  beforeEach(() => {
-    render(<VideoBox { ...props } />)
+  const props = {
+    video: 'video.mp4',
+    videoPreview: true
+  }
+
+  it('should render video preview when videoPreview is true', () => {
```
iva-stasia, line 10: It's not crucial here, but we can make the description clearer:
`should render video preview when a user doesn't provide a link to the video presentation`.
The same for the opposite scenario.

src/pages/tutor-profile/TutorProfile.jsx
D1nviz: `!loading && tutorVideo` will be false if loading is true, or if `response?.videoLink?.tutor is undefined, null, or an empty string.` This is because JavaScript treats `undefined, null, and "" as "falsy"` values.

tests/unit/components/video-box/VideoBox.spec.jsx
Mygal-Alex: I don't think it is necessary to check for two values. it is clear from the variable name that it is a video. if video !== null or empty string => execute action

src/components/video-box/VideoBox.tsx
```diff
@@ -1,20 +1,26 @@
...
+import videoImg from '~/assets/img/tutor-profile-page/presentationVideoImg.png'
+import VideoPlayer from '../video-player/VideoPlayer'
 
 interface VideoBoxProps {
   video: string
+  videoPreview: boolean
 }
 
-const VideoBox: FC<VideoBoxProps> = ({ video }) => {
+const VideoBox: FC<VideoBoxProps> = ({ video, videoPreview }) => {
```
iva-stasia, line 13: We should add a default value `videoPreview = true`. Otherwise, there is an issue with a video preview on the landing page:
image

Mygal-Alex: Oh, I hadn't noticed.
I'll fix it.

mhevyk:
Did you stick to figma design while implementing video player? Because I noticed, that there are small inconsistencies with mockup. For instance, border radius is not present on preview of your implementation, but exists on mockup.

Also on the mockup there is no mac-like wrapper around video or preview.

Your implementation:

image

Mockup: 

image

Figma mockup: https://www.figma.com/file/liJZDYhUnDP15Pi9bipDcv/Space2Study-(Students)?type=design&node-id=2749-59434&mode=design&t=BDbMYy8pDINkCg0y-0

src/pages/tutor-profile/TutorProfile.jsx
```diff
@@ -41,6 +41,13 @@ const TutorProfile = () => {
...
+      setVideoPreview(false)
+    }
+  }, [loading, response])
+
```
mhevyk, line 50: `useEffect` and state for storing video preview is redundant, because it is derived state. Please remove `useState` and `useEffect` for this. I suggest refactoring as following:

```suggestion
<VideoPresentation
    video={response?.videoLink?.[userRole]}
    videoPreview={loading || !tutorVideo}
 />
```

src/components/video-box/VideoBox.tsx
```diff
@@ -1,20 +1,26 @@
...
+    <VideoPlayer video={video} />
+  )
+
   return (
     <Box data-testid='videoBox' sx={styles.videoBox}>
       <Box alt='Title bar' component='img' src={titleBar} />
-      <Box sx={styles.videoBg}>
-        <Box alt='Video' component='img' src={video} sx={styles.video} />
-      </Box>
+      <Box sx={styles.videoBg}>{videoBloc}</Box>
```
mhevyk, line 23: Please refactor `videoBloc` name

src/containers/tutor-profile/video-presentation/VideoPresentation.jsx
```diff
@@ -2,17 +2,15 @@ import { useTranslation } from 'react-i18next'
 import Box from '@mui/material/Box'
 import Typography from '@mui/material/Typography'
 import VideoBox from '~/components/video-box/VideoBox'
-import videoImg from '~/assets/img/tutor-profile-page/presentationVideoImg.png'
 
-const VideoPresentation = () => {
+const VideoPresentation = (props) => {
```
mhevyk, line 6: Why did you decide to use `props` and not destructuring them? I also suggest writing this component in typescript. I hope it won't be very difficult.

mhevyk (changes requested):

src/containers/tutor-profile/video-presentation/VideoPresentation.jsx
asya982, line 6: Maybe you convert this component to Typescript, since is not clear what should be in props

src/pages/tutor-profile/TutorProfile.jsx
mhevyk: Definitely!

src/containers/tutor-profile/video-presentation/VideoPresentation.jsx
Mygal-Alex: I de-structured the props.
As for the component switching to typscript - this does not apply to my task. a separate ishyu should be described for this one.

src/components/video-box/VideoBox.tsx
Mygal-Alex: I replaced it with videoBlockContent.

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: I don't think it's redundant. 
this implementation is necessary because of the different behavior of mock implication. at 
Tutor profile and student profile 

Also in the code you suggested there is no check for the value of an empty string.
When such data arrives, the video player is displayed incorrectly.

src/components/video-box/VideoBox.tsx
Mygal-Alex: fixed

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: we need to check for both of these values

src/pages/tutor-profile/TutorProfile.jsx
mhevyk: I think you can add checks for student into condition, but extra hook usage is not good

src/components/video-box/VideoBox.tsx
```diff
@@ -1,20 +1,26 @@
 import { FC } from 'react'
 import Box from '@mui/material/Box'
-
 import titleBar from '~/assets/img/guest-home-page/title-bar.png'
 import { styles } from '~/components/video-box/VideoBox.styles'
+import VideoPlayer from '../video-player/VideoPlayer'
```
dmtrth25, line 5: use `aliases` for imports 
single code style

src/components/video-box/VideoBox.tsx
```diff
@@ -1,20 +1,26 @@
...
 interface VideoBoxProps {
   video: string
+  videoPreview: boolean
+  videoMock: string
 }
 
-const VideoBox: FC<VideoBoxProps> = ({ video }) => {
+const VideoBox: FC<VideoBoxProps> = ({ video, videoPreview, videoMock }) => {
+  const videoBlockContent = videoPreview ? (
+    <Box alt='Video' component='img' src={videoMock} sx={styles.video} />
```
dmtrth25, line 15: use `ComponentEnum`
For `component` and also for other code properties where necessary

src/pages/tutor-profile/TutorProfile.jsx
D1nviz:
```
const loading = false;
let tutorVideo;
```
// It just console.log, but if you bring this code to condition it will return true or false
// Case 1: tutorVideo is undefined
```
tutorVideo = undefined;
console.log(!loading && tutorVideo); // Output: false
```

// Case 2: tutorVideo is null
```
tutorVideo = null;
console.log(!loading && tutorVideo); // Output: false
```

// Case 3: tutorVideo is an empty string
```
tutorVideo = '';
console.log(!loading && tutorVideo); // Output: false
```

// Case 4: tutorVideo is a non-empty string
```
tutorVideo = 'some value';
console.log(!loading && tutorVideo); // Output: 'some value'
```

src/components/video-player/VideoPlayer.styles.ts
```diff
@@ -0,0 +1,11 @@
+export const styles = {
+  playerWrapper: {
+    position: 'relative',
+    paddingTop: '56.25%'
```
dmtrth25, line 4: we use short convention naming `pt`
I'm suggest round down/up `56.25%` what do you think ?

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: I understand, but there's a downside. 
a userfect can protect us from race conditions.
Also the realization that we have now is quite readable. 
We can see both the required checks that a string must pass and under what condition it should display the video preview.

src/pages/tutor-profile/TutorProfile.jsx
```diff
@@ -54,7 +62,13 @@ const TutorProfile = () => {
       <ProfileInfo myRole={userRole} userData={response} />
       <CompleteProfileBlock data={response} profileItems={profileItems} />
       <AboutTutorBlock />
-      {shouldShowPresentation && <VideoPresentation />}
+      {shouldShowPresentation && (
```
dmtrth25, line 65: for good practice it should be in the new component and use here. Can do that or pass it into variable and render `{showPresentation}` smth like that

src/pages/tutor-profile/TutorProfile.jsx
mhevyk: Can you explain in more details why your solution is better than proposed one? Because I don't get a point why it protects us from race condition

tests/unit/components/video-box/VideoBox.spec.jsx
```diff
@@ -1,19 +1,27 @@
...
 import VideoBox from '~/components/video-box/VideoBox'
 
-const props = {
-  video: 'video.mp4'
-}
-
 describe('VideoBox component', () => {
-  beforeEach(() => {
-    render(<VideoBox { ...props } />)
+  const props = {
```
dmtrth25, line 5: all mocked data we extract out of the `describe` block and out of the `it` tests and write it above

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: I meant it in terms of my code. 
image
if we use this setter (thereby explicitly specifying the state when we should not show video previews)
without using the userfect 
we get this result 
image

mhevyk: You should remove `useState`, it is also redundant here. Thats why you get an error with infinite rerenders

Mygal-Alex: image
fixed

src/components/video-box/VideoBox.tsx
Mygal-Alex: image
right now it looks like this

src/components/video-player/VideoPlayer.styles.ts
Mygal-Alex: we can't give it up.
because we need to keep the aspect ratio 16:9.
Let's take a 1280 x 720 player as an example.
720/1280 = 0.5625
0.5625 * 100 = 56.25

tests/unit/components/video-box/VideoBox.spec.jsx
Mygal-Alex: image
fixed

src/pages/tutor-profile/TutorProfile.jsx
mhevyk: I also suggest renaming `showPresentation` to something like `VideoPresentationComponent` to indicate that it is a `jsx` component and not a condition

src/components/video-player/VideoPlayer.styles.ts
D1nviz: could we use [aspect ratio](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio) property instead?

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: image
fixed

src/components/video-box/VideoBox.tsx
Mygal-Alex: image
fixed

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: image
fixed

src/components/video-player/VideoPlayer.styles.ts
Mygal-Alex: image
fixed

src/pages/tutor-profile/TutorProfile.jsx
D1nviz: good job!

src/components/video-player/VideoPlayer.styles.ts
D1nviz: Cool! It looks cleaner now

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: fixed

mhevyk: Good job!

mhevyk (approved):

D1nviz (approved):

iva-stasia (approved):

asya982 (approved):

GeorgeShvab (approved):

src/pages/tutor-profile/TutorProfile.jsx
byavd: what a valid falsy value are you expecting here, that just `... && tutorVideo` is not enough?

src/pages/tutor-profile/TutorProfile.jsx
```diff
@@ -41,20 +41,43 @@ const TutorProfile = () => {
     defaultResponse: defaultResponses.array
   })
 
+  useEffect(() => {
+    if (
+      !loading &&
+      response &&
+      response.videoLink &&
+      response.videoLink.tutor !== null &&
+      response.videoLink.tutor !== ''
```
byavd: what a valid falsy value are you expecting here, that just `... && response.videoLink.tutortutorVideo `is not enough?

byavd (changes requested):

src/pages/tutor-profile/TutorProfile.jsx
Mygal-Alex: this implementation is outdated.
Now the check is implemented differently

byavd (approved):

PR #1813, opened 2024-04-17
Refactor FileUploader to emit action when the same file is added twice in a row

D1nviz:
The `onChange` event for an input of type file doesn't trigger when the same file is selected twice in a row because the value of the input doesn't actually change. The browser doesn't consider selecting the same file again as a change in value, so it doesn't trigger the `onChange` event.

So i considered to reset `event.target.value` after we adding a file to resolve this problem

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/118522791/c5b3dea1-aabf-439f-986e-349e23a54843

### DIFF
--- a/src/hooks/use-upload.tsx
+++ b/src/hooks/use-upload.tsx
@@ -37,6 +37,7 @@ const useUpload = ({ files, validationData, emitter }: UseUpload) => {
     const error = filesValidation(newFiles, validationData) ?? ''
     const filesForEmitter = error ? files : newFiles
     emitter({ files: filesForEmitter, error })
+    event.target.value = ''
   }
 
   const deleteFile = (file: File) => {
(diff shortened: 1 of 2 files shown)


mhevyk (approved):

asya982 (approved):

t-valentyna (approved):

GeorgeShvab (approved):

src/hooks/use-upload.tsx
```diff
@@ -37,6 +37,7 @@ const useUpload = ({ files, validationData, emitter }: UseUpload) => {
     const error = filesValidation(newFiles, validationData) ?? ''
     const filesForEmitter = error ? files : newFiles
     emitter({ files: filesForEmitter, error })
+    event.target.value = ''
```
byavd, line 40: It is a bad practice to directly mutate function arguments. You should avoid this. I believe there should be a better solution for your problem

byavd (changes requested):

src/hooks/use-upload.tsx
D1nviz: Used ref istead. After button click we resetting previous data to solve this problem

byavd (approved):

PR #1814, opened 2024-04-18
Сonvert use select hook into TS/1804

OliaHrytsak:
- useSelect hook was converted into TS
- According to these changes, also changes were made in use-select.spec.jsx and in component that use this hook


src/hooks/table/use-select.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  const createSelectAllHandler =
+    (items: Item[]) => (e: ChangeEvent<HTMLInputElement>) =>
+      handleSelectAllClick(e, items)
+
+  const handleSelectClick = (id: string) => {
+    const selectedIndex = selected.indexOf(id)
+    let newSelected: string[] = []
+
```
byavd, line 44:
```suggestion
  const newSelected = selected.includes(id)
    ? selected.filter(selectedId => selectedId !== id)
    : [...selected, id];

  setSelected(newSelected);
```

src/hooks/table/use-select.tsx
```diff
@@ -0,0 +1,65 @@
...
+    items: Item[]
+  ) => (e: ChangeEvent<HTMLInputElement>) => void
+  handleSelectClick: (id: string) => void
+}
+
+const useSelect = ({
+  initialSelect = []
+}: {
+  initialSelect?: string[]
+}): SelectHook => {
```
mhevyk, line 21: The more readable way is to type things here is to create separate interfaces for arguments and return type. I suggest the following:

```jsx
interface UseSelectInput {
    // ... your fields
}

interface UseSelectOutput {
    // ... your fields
}

const useSelect = ({ initialSelect = [] }: UseSelectInput): UseSelectOutput // ... other code
```

You can find similar approach inside `useForm` or `useFilterQuery` hooks

src/hooks/table/use-select.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const clearSelected = useCallback(() => setSelected([]), [setSelected])
+
+  const handleSelectAllClick = (
+    e: ChangeEvent<HTMLInputElement>,
+    items: Item[]
+  ) => {
+    if (e.target.checked) {
+      const newSelected = items.map((item) => item._id)
+      return setSelected(newSelected)
```
mhevyk, line 32: Maybe it would be better to move `return` statement to a new line, because it is a bit confusing whether we return something from `setSelected` or we just quitung function `handleSelectAllClick`

src/hooks/table/use-select.tsx
OliaHrytsak: Thanck uoy for your advice, but I think that in this case it is not necessary and will only make readability more difficult

src/containers/add-resources/AddResources.tsx
```diff
@@ -109,7 +109,11 @@ const AddResources = <T extends CourseResources | Question>({
 
         const categoryMatch =
           selectedCategories.length === 0 ||
-          selectedCategories.includes(String(item.category?._id || null))
+          selectedCategories.includes(
+            String(
+              (typeof item.category !== 'string' && item.category?._id) || null
+            )
+          )
```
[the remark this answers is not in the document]
OliaHrytsak: Fixed! Thanks

src/hooks/table/use-select.tsx
OliaHrytsak: Cool! Thanks!

src/hooks/table/use-select.tsx
mhevyk: It definitely will improve readability in case adding more props in the future. And by the way it is much cleaner way to declare types for function

D1nviz (approved):

asya982 (approved):

src/hooks/table/use-select.tsx
mhevyk: No, i meant like following:

```jsx
setSelected(newSelected)
return
```

mhevyk (approved):

PR #1815, opened 2024-04-18
Fixed category duplication

GeorgeShvab:
- I added check if string contains only spaces in `emptyField` validation. I added it so it is impossible to create a category with name consisting only of spaces. It will validate any other fields that have validation for empty values, I guess that would be good for most cases. In general, text without any characters and text consisting only of spaces are both empty from user's perspective. Also, I added error message for it, case default empty message might be confusing in this case.
 
- Added spaces removal at the end and start for category name value before sending to server, so there would be no confusion like this: "category" and "category   ", which look identical to user.

- I also replaced `AsyncAppAutocomplete` with `AppAutoComplete` inside of `CategoryDropdown`. I did it because `AsyncAppAutocomplete` takes service as an argument and makes request inside of it, in this case it gets categories names. But the same data is also needed in another child of `CategoryDrodpown` (`AddCategoriesModal`), so to avoid making two identical requests, I added `useAxios` request for categories names inside of `CategoryDropdown`, which then passes it to children. As I used `AppAutocomplete`, I also added types to it.
 
- Also, added `useAxios` request for categories names inside of `categoriesContainer` as it also has two children that use categories names.
 
- There were some mistakes regarding spaces in translations, so prettier fixed them in those files where I added translations for new validation.

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/62070431/2eebe46c-3482-40b5-90b9-03cf273644c5


src/containers/category-dropdown/CategoryDropdown.tsx
```diff
@@ -81,9 +85,9 @@ const CategoryDropdown = ({
         })
       })
 
-      setIsFetched(false)
+      void fetchAllCategoriesNames()
```
asya982, line 88: what does this line mean?

GeorgeShvab: `setIsFetched` was responsible for indicating whether it needed to fetch categories. When categories were fetched, this state was changed to true. When new category was created, it switched to false (that is the line you selected). There was this logic, because categories were fetched on focus. This state existed so categories wouldn't be fetched on every focus, but only on those before which new category was created (and there were no fetches after category was created). I slightly changed this logic to fetch categories each time after new category is created (I added call to do this in line you selected), just inside `onResponse` callback. I think if user creates new category inside of this dropdown, he will definitely want it to appear in this list, so there is no sense to wait for focus on dropdown.

asya982: Thank you so much for this explanation, very informative, totally agree with your point! 
But i was curios `void fetchAllCategoriesNames()` about this line, never saw `void` before function call

asya982 (approved):

src/components/location-selection-inputs/LocationSelectionInputs.tsx
```diff
@@ -93,7 +93,7 @@ const LocationSelectionInputs = <
       <AppAutoComplete
         fullWidth
         loading={loadingCountries}
-        onChange={onChangeCountry}
+        onChange={void onChangeCountry}
```
byavd, line 96: What the point of using `void` operator here and other places?

src/containers/my-resources/add-categories-modal/AddCategoriesModal.constants.ts
```diff
@@ -4,11 +4,17 @@ export const initialValues = {
...
+const valueNotIn = (value: string, disallowedValues: string[]) =>
+  disallowedValues.includes(value)
+    ? 'myResourcesPage.categories.categoryAlreadyExistsError'
+    : undefined
+
+export const validations = (disallowedValues: string[]) => ({
+  name: (value: string) => {
+    const trimmedValue = value.trim()
+    return (
+      emptyField(value, undefined, textField(2, 35)(trimmedValue)) ||
```
byavd, line 16: Instead of passing `undefined` to you function, you can refactor your function to expect object as an argument and then just not passing that key on function call. Therefor that key will be undefined

src/containers/my-resources/categories-container/CategoriesContainer.constansts.tsx
```diff
@@ -5,11 +5,21 @@ import RenameInput from '~/containers/my-resources/rename-input/RenameInput'
 import { getFormattedDate } from '~/utils/helper-functions'
 import { Categories, TableColumn, SortEnum, RemoveColumnRules } from '~/types'
 import { styles } from '~/containers/my-resources/categories-container/CategoriesContainer.style'
+import { emptyField, textField } from '~/utils/validations/common'
+
+export const validation = (disallowedValues: string[]) => (value: string) => {
+  const trimmedValue = value.trim()
+  return Boolean(
+    emptyField(value, undefined, textField(2, 35)(trimmedValue)) ||
```
byavd, line 13: same

byavd (changes requested):

src/containers/category-dropdown/CategoryDropdown.tsx
GeorgeShvab: I did it to overcome lint error that occurs when trying to call func that returns promise without await: `Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the void operator  @typescript-eslint/no-floating-promises`. Since Bohdan also mentioned it, I will try to overcome it another way.

src/containers/my-resources/add-categories-modal/AddCategoriesModal.constants.ts
GeorgeShvab: Refactored emptyField, made object parameter, changed arguments in places where it is used

byavd (approved):

D1nviz (approved):

OliaHrytsak (approved):

PR #1835, opened 2024-04-23
Fixed edit profile button

GeorgeShvab:
I created a new component for this task, so we won't need to pass `component={Link}` or wrap `<IconButton>` in `<Link>` each time we want to make icon button to behave as a link.

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/62070431/4bbe5d56-4dca-4b39-92c5-283be76bc3e8

### DIFF
--- a/src/components/app-icon-button/AppIconButton.tsx
+++ b/src/components/app-icon-button/AppIconButton.tsx
@@ -1,0 +1,12 @@
+import { IconButton, IconButtonProps } from '@mui/material'
+import { FC } from 'react'
+import { Link, LinkProps } from 'react-router-dom'
+import { ComponentEnum } from '~/types'
+
+type AppIconButtonProps = IconButtonProps & Partial<LinkProps>
+
+const AppIconButton: FC<AppIconButtonProps> = ({ to, ...props }) => (
+  <IconButton component={to ? Link : ComponentEnum.Button} to={to} {...props} />
+)
+
+export default AppIconButton
--- a/src/containers/tutor-profile/profile-info/ProfileInfo.jsx
+++ b/src/containers/tutor-profile/profile-info/ProfileInfo.jsx
@@ -3,7 +3,7 @@ import { useTranslation } from 'react-i18next'
 
 import Box from '@mui/material/Box'
 import Button from '@mui/material/Button'
-import IconButton from '@mui/material/IconButton'
+import AppIconButton from '~/components/app-icon-button/AppIconButton'
 import EditOutlinedIcon from '@mui/icons-material/EditOutlined'
 import CopyRoundedIcon from '@mui/icons-material/ContentCopyRounded'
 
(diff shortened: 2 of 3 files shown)


src/components/app-icon-button/AppIconButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+const AppIconButton: FC<AppIconButtonProps> = ({ children, to, ...props }) => {
+  if (to) {
+    return (
+      <IconButton component={Link} to={to} {...props}>
+        {children}
+      </IconButton>
+    )
+  }
+
+  return <IconButton {...props}>{children}</IconButton>
```
mhevyk, line 18: You can omit children in props destructuring and refactor here like following:

```suggestion
  return <IconButton {...props} />
```

src/components/app-icon-button/AppIconButton.tsx
```diff
@@ -0,0 +1,21 @@
+import { IconButton, IconButtonProps } from '@mui/material'
+import { FC } from 'react'
+import { Link, To } from 'react-router-dom'
+
+interface AppIconButtonProps extends IconButtonProps {
+  to?: To
```
mhevyk, line 6: How do you think, do we need to predict passing other link props here also because we are using only `to` prop here?
`react-router-dom` has following own props in `Link` (from docs):

```jsx
export interface LinkProps extends Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, "href"> {
    reloadDocument?: boolean;
    replace?: boolean;
    state?: any;
    preventScrollReset?: boolean;
    relative?: RelativeRoutingType;
    to: To;
    unstable_viewTransition?: boolean;
}
```

And do we need to take into account `NavLink` possible usage here or it is redundant?

src/components/app-icon-button/AppIconButton.tsx
GeorgeShvab: Refactored, thanks for suggestion.

src/components/app-icon-button/AppIconButton.tsx
GeorgeShvab: I added `LinkProps` to props type, so we can pass any `react-router-link` `Link` props. As for `NavLink`, it is not used anywhere in the project now, I don't think there would be many places where it can be used with Icon.

mhevyk (approved):

src/components/app-icon-button/AppIconButton.tsx
mhevyk: Now it became better!

OliaHrytsak (approved):

t-valentyna (approved):

asya982 (approved):

src/components/app-icon-button/AppIconButton.tsx
```diff
@@ -0,0 +1,15 @@
+import { IconButton, IconButtonProps } from '@mui/material'
+import { FC } from 'react'
+import { Link, LinkProps } from 'react-router-dom'
+
+type AppIconButtonProps = IconButtonProps & Partial<LinkProps>
+
+const AppIconButton: FC<AppIconButtonProps> = ({ to, ...props }) => {
+  if (to) {
+    return <IconButton component={Link} to={to} {...props} />
```
D1nviz, line 9: Maybe would be better do something like this
```
const AppIconButton: FC<AppIconButtonProps> = ({ to, ...props }) => (
  <IconButton component={to ? Link : 'button'} to={to} {...props} />
)
```

GeorgeShvab: Thanks, used your approach so repeating `<IconButton/>` can be avoided

iva-stasia (approved):

D1nviz (approved):

dmtrth25 (approved):

dmtrth25:
George. Don't forget to add our project on the right side `Space2Study` to control process
image

byavdoshniak (approved):

PR #1837, opened 2024-04-24
Add user photo instead of user menu icon/1823

OliaHrytsak:
In this pull request:
- Created new component AccountIcon (in the folder navigation-icons)(there was no posibility to add this element to exisisting file with NavigationIcons.constants)
- Added tests to this implementation
- Removed previous MUI element from NavigationIcons.constants
- Fixed tests for NavBar and UserIcons component
- Added immediate reload of the page after clicking on 'Update profile' button for changin the Avatar without user intervention

image

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/127191734/600bdfde-088c-4d79-a518-7f99c3c1c3a4

### DIFF
--- a/src/containers/edit-profile/profile-tab/profile-tab-inner-container/ProfileTabInnerContainer.tsx
+++ b/src/containers/edit-profile/profile-tab/profile-tab-inner-container/ProfileTabInnerContainer.tsx
@@ -20,11 +20,6 @@ const ProfileTabInnerContainer: FC<ProfileTabProps> = ({ user }) => {
     profileData.generalData.errors &&
     Object.values(profileData.generalData.errors).some((error) => error)
 
-  const handleUpdateProfile = () => {
-    handleSubmit(profileData.generalData.data)
-    window.location.reload()
-  }
-
   return (
     <Box sx={styles.profileInnerContainer}>
       <Box sx={styles.root}>
@@ -39,7 +34,7 @@ const ProfileTabInnerContainer: FC<ProfileTabProps> = ({ user }) => {
       <AppButton
         disabled={hasError}
         loading={loading}
-        onClick={handleUpdateProfile}
+        onClick={() => handleSubmit(profileData.generalData.data)}
         size={SizeEnum.ExtraLarge}
         sx={styles.updateProfileBtn}
         variant={ButtonVariantEnum.Contained}
--- a/src/containers/edit-profile/profile-tab/profile-tab-inner-container/ProfileTabInnerContainer.tsx
+++ b/src/containers/edit-profile/profile-tab/profile-tab-inner-container/ProfileTabInnerContainer.tsx
@@ -20,6 +20,11 @@ const ProfileTabInnerContainer: FC<ProfileTabProps> = ({ user }) => {
     profileData.generalData.errors &&
     Object.values(profileData.generalData.errors).some((error) => error)
 
+  const handleUpdateProfile = () => {
+    handleSubmit(profileData.generalData.data)
+    window.location.reload()
+  }
+
   return (
     <Box sx={styles.profileInnerContainer}>
       <Box sx={styles.root}>
@@ -34,7 +39,7 @@ const ProfileTabInnerContainer: FC<ProfileTabProps> = ({ user }) => {
       <AppButton
         disabled={hasError}
         loading={loading}
-        onClick={() => handleSubmit(profileData.generalData.data)}
+        onClick={handleUpdateProfile}
         size={SizeEnum.ExtraLarge}
         sx={styles.updateProfileBtn}
         variant={ButtonVariantEnum.Contained}
--- a/src/containers/navigation-icons/AccountIcon.tsx
+++ b/src/containers/navigation-icons/AccountIcon.tsx
@@ -27,31 +27,38 @@ const AccountIcon: FC<AccountIconProps> = ({ openMenu }) => {
     [userId, userRole]
   )
 
-  const { loading, response } = useAxios<UserResponse>({
+  const {
+    loading,
+    response: { photo, firstName, lastName }
+  } = useAxios<UserResponse>({
     service: getUserData,
     fetchOnMount: true,
     defaultResponse: defaultResponses.object as UserResponse
   })
 
-  const photoUrl =
-    response.photo &&
-    `${import.meta.env.VITE_APP_IMG_USER_URL}${response.photo}`
+  if (loading) {
+    return <Avatar sx={styles.accountIcon} />
+  }
 
-  const userNameForAvatar = loading
-    ? ''
-    : `${response.firstName?.slice(0, 1) || ''}${
-        response.lastName?.slice(0, 1) || ''
-      }`
+  let userPhoto = ''
+  if (!loading && photo) {
+    userPhoto = `${import.meta.env.VITE_APP_IMG_USER_URL}${photo}`
+  }
+
+  let userNameInitials = ''
+  if (!loading && firstName && lastName) {
+    userNameInitials = `${firstName[0] || ''}${lastName[0] || ''}`
+  }
 
   return (
     <Tooltip title={t('iconsTooltip.account')}>
       <Avatar
         alt='User'
         onClick={openMenu}
-        src={photoUrl}
+        src={userPhoto}
         sx={styles.accountIcon}
       >
-        {userNameForAvatar}
+        {userNameInitials}
       </Avatar>
     </Tooltip>
   )
--- a/src/containers/navigation-icons/AccountIcon.tsx
+++ b/src/containers/navigation-icons/AccountIcon.tsx
@@ -1,4 +1,4 @@
-import { useCallback, FC } from 'react'
+import { useCallback, FC, MouseEvent } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useAppSelector } from '~/hooks/use-redux'
 import { AxiosResponse } from 'axios'
@@ -15,7 +15,7 @@ import { styles } from '~/containers/navigation-icons/NavigationIcons.styles'
 import { UserResponse, UserRole } from '~/types'
 
 interface AccountIconProps {
-  openMenu: () => void
+  openMenu: (event: MouseEvent) => void
 }
 
 const AccountIcon: FC<AccountIconProps> = ({ openMenu }) => {
--- a/src/containers/navigation-icons/AccountIcon.tsx
+++ b/src/containers/navigation-icons/AccountIcon.tsx
@@ -39,7 +39,7 @@ const AccountIcon: FC<AccountIconProps> = ({ openMenu }) => {
 
   const userNameForAvatar = loading
     ? ''
-    : response.firstName.charAt(0) + '' + response.lastName.charAt(0)
+    : `${response.firstName.slice(0, 1)}${response.lastName.slice(0, 1)}`
 
   return (
     <Tooltip title={t('iconsTooltip.account')}>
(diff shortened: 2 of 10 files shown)


src/containers/navigation-icons/AccountIcon.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const photoUrl =
+    response.photo &&
+    `${import.meta.env.VITE_APP_IMG_USER_URL}${response.photo}`
+
+  const userNameForAvatar = loading
+    ? ''
+    : `${response.firstName?.slice(0, 1) || ''}${
+        response.lastName?.slice(0, 1) || ''
+      }`
```
mhevyk, line 44: I think using `if` check would be more cleaner solution here. I suggest refactoring like this:

```suggestion
let usernameInitials = "";
if (loading && response) {
    usernameInitials = `${response.firstName[0] ?? ""}${response.lastName[0] ?? ""}`
}
```

I am not sure about a need to use `?? ""` here. Please double check it.

src/containers/navigation-icons/AccountIcon.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const { loading, response } = useAxios<UserResponse>({
+    service: getUserData,
+    fetchOnMount: true,
+    defaultResponse: defaultResponses.object as UserResponse
+  })
+
+  const photoUrl =
+    response.photo &&
+    `${import.meta.env.VITE_APP_IMG_USER_URL}${response.photo}`
```
mhevyk, line 38: Same thing here. I think `if` check will be more readable is this case

src/containers/navigation-icons/AccountIcon.tsx
mhevyk: Why did you decide not to use `loading` variable to show spinner for example? Maybe it would be better to show some fallback when image and user data are loading

tests/unit/containers/navigation-icons/AccountIcon.spec.jsx
```diff
@@ -0,0 +1,34 @@
+import { fireEvent, screen, waitFor } from '@testing-library/react'
+import { expect } from 'vitest'
```
mhevyk, line 2: I think it is redundant here

mhevyk:
As for me, font size should be a bit smaller for better displayment. What do you think?

image

src/containers/navigation-icons/AccountIcon.tsx
```diff
@@ -0,0 +1,60 @@
...
+const AccountIcon: FC<AccountIconProps> = ({ openMenu }) => {
+  const { t } = useTranslation()
+  const { userId, userRole } = useAppSelector((state) => state.appMain)
+
+  const getUserData: () => Promise<AxiosResponse<UserResponse>> = useCallback(
+    () => userService.getUserById(userId, userRole as UserRole),
+    [userId, userRole]
+  )
+
+  const { loading, response } = useAxios<UserResponse>({
```
asya982, line 30:
```suggestion
  const { loading, response: { photo, firstName, LastName } } = useAxios<UserResponse>({
```

And then you can use them directly + in lines 42-43, you can make them more readable by writting
`firstName[0] || ''` && `lastName[0] || ''`
What do you think about this approach?

OliaHrytsak: Thank you for this advice! Realy looks much better and more readable)

OliaHrytsak:
> As for me, font size should be a bit smaller for better displayment. What do you think?
> 
> image

Fixed) Thank you!
                                    image

tests/unit/containers/navigation-icons/AccountIcon.spec.jsx
OliaHrytsak: Fixed!

src/containers/navigation-icons/AccountIcon.tsx
OliaHrytsak: Thank you! Realy using the nullish coalescing operator (??) generally considered better for handling default values in TypeScript

I've desided to use default Avatar when loading. Thanks!

image

src/containers/edit-profile/profile-tab/profile-tab-inner-container/ProfileTabInnerContainer.tsx
```diff
@@ -20,6 +20,11 @@ const ProfileTabInnerContainer: FC<ProfileTabProps> = ({ user }) => {
     profileData.generalData.errors &&
     Object.values(profileData.generalData.errors).some((error) => error)
 
+  const handleUpdateProfile = () => {
+    handleSubmit(profileData.generalData.data)
+    window.location.reload()
```
iva-stasia, line 25: I guess it's better to move this into the `handleResponse` function in the `useUpdateUser` hook. Because we need to reload only if the user info was updated successfully.

Also, do we need to reload the page if only text fields were changed?

One last thought: I think we should either add a short delay before reloading to give users a chance to read the message from the snackbar, or get rid of it completely. Because right now we don't have enough time to read it, and it's a bit confusing.

OliaHrytsak: Thanks for comment. Fixed! Now page reload only if photo changed and after short delay.

src/containers/navigation-icons/AccountIcon.tsx
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  let userPhoto = ''
+  if (!loading && photo) {
+    userPhoto = `${import.meta.env.VITE_APP_IMG_USER_URL}${photo}`
+  }
+
+  let userNameInitials = ''
+  if (!loading && firstName && lastName) {
+    userNameInitials = `${firstName[0] || ''}${lastName[0] || ''}`
```
asya982, line 50:
```suggestion
    userNameInitials = `${firstName[0]}${lastName[0]}`
```
since you already check for availability of firstName and lastName, maybe you don't need to give an alternative anymore

D1nviz (approved):

mhevyk (approved):

Mav-Ivan (approved):

src/containers/navigation-icons/AccountIcon.tsx
OliaHrytsak: fixed

src/containers/navigation-icons/AccountIcon.tsx
OliaHrytsak: fixed)

iva-stasia (approved):

asya982 (approved):

dmtrth25 (approved):

src/containers/navigation-icons/AccountIcon.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  let userNameInitials = ''
+  if (!loading && firstName && lastName) {
+    userNameInitials = `${firstName[0]}${lastName[0]}`
+  }
+
+  return (
+    <Tooltip arrow title={t('iconsTooltip.account')}>
+      <Avatar
+        alt='User'
```
byavdoshniak, line 56:
```suggestion
        alt='User Avatar'
```

src/hooks/use-update-user.tsx
```diff
@@ -18,6 +18,7 @@ const useUpdateUser = (userId: string) => {
       severity: snackbarVariants.success,
       message: 'editProfilePage.profile.successMessage'
     })
+    window.location.reload()
```
byavdoshniak, line 21: use `react-router` tools to reload the page

src/containers/navigation-icons/AccountIcon.tsx
OliaHrytsak: Fixed

src/hooks/use-update-user.tsx
OliaHrytsak: Thank you for advice! Fixed

src/containers/navigation-icons/AccountIcon.tsx
```diff
@@ -0,0 +1,67 @@
...
+    service: getUserData,
+    fetchOnMount: true,
+    defaultResponse: defaultResponses.object as UserResponse
+  })
+
+  if (loading) {
+    return <Avatar sx={styles.accountIcon} />
+  }
+
+  let userPhoto = ''
```
Tolik170, line 43: redundant variable, just pass `${import.meta.env.VITE_APP_IMG_USER_URL}${photo}` into src

src/containers/navigation-icons/AccountIcon.tsx
```diff
@@ -0,0 +1,67 @@
...
+  if (!loading && photo) {
+    userPhoto = `${import.meta.env.VITE_APP_IMG_USER_URL}${photo}`
+  }
+
+  let userNameInitials = ''
```
Tolik170, line 48: same for this one

src/containers/navigation-icons/NavigationIcons.styles.ts
```diff
@@ -14,5 +15,14 @@ export const styles = {
...
   },
-  studentIcons: { ...hideOnMobile, color: 'primary.900' }
+  studentIcons: {
+    ...hideOnMobile,
+    color: 'primary.900'
+  },
+  accountIcon: {
+    ...hideOnMobile,
+    color: 'primary.900',
+    ml: { sm: '12px', md: '7px', lg: '12px' },
```
Tolik170, line 25: Why you can not use 12px for `md` too and what it will have for `xs`

src/containers/navigation-icons/NavigationIcons.styles.ts
```diff
@@ -14,5 +15,14 @@ export const styles = {
...
     ml: '12px'
   },
-  studentIcons: { ...hideOnMobile, color: 'primary.900' }
+  studentIcons: {
+    ...hideOnMobile,
+    color: 'primary.900'
+  },
+  accountIcon: {
+    ...hideOnMobile,
+    color: 'primary.900',
```
Tolik170, line 24: can you also move color property to hideOnMobile?

src/hooks/use-update-user.tsx
```diff
@@ -18,6 +21,7 @@ const useUpdateUser = (userId: string) => {
       severity: snackbarVariants.success,
       message: 'editProfilePage.profile.successMessage'
     })
+    navigate(0)
```
Tolik170, line 24: where it will navigate?

OliaHrytsak: It will reload the page after successful photo update.
It was advice of the expert to use router for reload

src/containers/navigation-icons/NavigationIcons.styles.ts
OliaHrytsak: Moved. Thank you

src/containers/navigation-icons/NavigationIcons.styles.ts
OliaHrytsak: It can be! Changed. 
It dosnt display on sm-screen, because of that, I think,  we don't need ml for xs

src/containers/navigation-icons/AccountIcon.tsx
OliaHrytsak: Fixed

src/containers/navigation-icons/AccountIcon.tsx
OliaHrytsak: It was made for increasing readability of code.
But I didn't consider that as redundant variable(((
thank you. Fixed

Tolik170: You use it only in one place and it isn`t a very long statement, also, defining variables with let keyword isn`t a good practice at all

Tolik170 (approved):

byavdoshniak (approved):

Issue #1841, opened 2024-04-25
(SP: ) Delete mui button styles from theme


mhevyk:
Delete mui button styles from theme and make sure ui for buttons did not break after changes

PR #1842, opened 2024-04-25
Feature/1807/design system button

mhevyk:
Implemented Button in design system according to [Figma design](https://www.figma.com/file/liJZDYhUnDP15Pi9bipDcv/Space2Study-(Students)?type=design&node-id=3-44537&mode=design&t=XbODb8Dk9rJdrHwN-0)

### Usage:

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/109922101/c92e154e-032f-4cbf-b34c-4b5b6bc9dd61

Now buttons look like following:

image

### Important notes:

1. Removed `Outlined` button variant as it was not used in project at all (also not planned future designs in figma).
2. Actual styles from mui theme were not deleted in order to not to break all existing ui for buttons and button like elements. I have already created issue for it #1841.
3. Added separate context to avoid `eslint` nesting error. We can create an issue also for better context managing.
4. Added `scss` compilation step before running our commands because it relies on compiled `css` files


src/scss/src/atoms/Button.scss
```diff
@@ -0,0 +1,133 @@
+@import '../foundation/mixins';
+@import '../foundation/colors';
+@import '../foundation/typography';
```
asya982, line 3: We do have file `_all.scss` isn't it better to import it here instead of 3 separate imports?

src/scss/src/atoms/Button.scss
```diff
@@ -0,0 +1,133 @@
...
+
+.s2s-button {
+  @extend %button-base;
```
asya982, line 6: Why do we need extends and below write the actual classes, why don't implement it here at once?
This styles will be used only here, it's seems unnessesary here, don't you think?

src/scss/src/atoms/Button.scss
mhevyk: Thanks for useful remark. Fixed

src/scss/src/atoms/Button.scss
mhevyk: You are right. I moved styles from `%button-base` into `s2s-button` directly. But I didn't move other subclasses with variants and sizes because it improves readability and unloads the component to avoid deep nesting.

iva-stasia (approved):

byavdoshniak (approved):

asya982 (approved):

D1nviz (approved):

PR #1845, opened 2024-04-26
Implemented pop-up on edit profile page

t-valentyna:
Implemented pop-up on edit profile page when leaving the Profile page with a half-filled form.

- Modal window appearance is achieved by using the `useBlocker` React Router hook.
- Since we decided to split tabs of the Profile tab, the unnecessary context is deleted and the Profile tab is refactored accordingly.
- Tests are added correspondingly.

The same logic will be added to other tabs when they are ready.

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/110097862/62fa10bb-9053-44ad-9f53-71943d0d9abc

Screenshot 2024-05-02 at 17 22 18

Screenshot 2024-05-02 at 16 11 02


byavdoshniak (approved):

Tolik170 (approved):

src/containers/edit-profile/profile-tab/profile-tab-form/ProfileTabForm.tsx
```diff
@@ -116,10 +93,7 @@ const ProfileGeneralTab: FC<ProfileTabProps> = ({ user }) => {
   }
 
   const handleRemovePhoto = () => {
-    handleProfileData({
-      ...updatedData,
-      photo: null
-    })
+    setPhoto(null)
```
GeorgeShvab, line 96: When user without photo clicks remove photo button, nothing changes for user, but the pop-up still appears after click on one of navigation buttons. The same when user without photo sets photo and without saving clicks remove photo button. Is that a desired behaviour? Maybe would be good to set photo to `null` only if the user has something inside of `user.photo`, and set it to initial photo state (`{ src: '', name: '' }` in `ProfileTab`) in other case? Something like this: `user.photo ? null : { src: '', name: '' }`.

src/containers/edit-profile/profile-tab/ProfileTab.tsx
```diff
@@ -1,16 +1,123 @@
-import { FC } from 'react'
-import { ProfileProvider } from '~/context/profile-context'
-import { ProfileTabProps } from '~/types'
-import ProfileTabInnerContainer from '~/containers/edit-profile/profile-tab/profile-tab-inner-container/ProfileTabInnerContainer'
-import { getProfileContextInitialValues } from '~/utils/get-profile-context-initial-values'
+import { FC, useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { Box } from '@mui/material'
```
mhevyk, line 3: Use default import for `Box` instead. From [docs](https://mui.com/material-ui/guides/minimizing-bundle-size/):

image

src/containers/edit-profile/profile-tab/ProfileTab.tsx
```diff
@@ -1,16 +1,123 @@
...
+    errors
+  } = useForm<EditProfileForm>({
+    initialValues,
+    validations
+  })
+
+  const isPhotoChanged = photo === null || Boolean(photo.src)
+  const isDataChanged = isDirty || isPhotoChanged
+
+  const blocker = useBlocker(isDataChanged)
```
mhevyk, line 57: Very interesting solution! I have never heard about this hook before

asya982 (approved):

src/containers/edit-profile/profile-tab/profile-tab-form/ProfileTabForm.tsx
iva-stasia: It's not a desired behavior indeed, I missed this case. Changed it based on your suggestion, thanks a lot.

src/containers/edit-profile/profile-tab/ProfileTab.tsx
iva-stasia: Thanks for noticing, changed ✅

src/containers/edit-profile/profile-tab/ProfileTab.tsx
iva-stasia: I used it for the first time as well, and it turned out to be quite useful in cases where we need to block navigation from a particular page for some reason.

D1nviz (approved):

src/containers/edit-profile/profile-tab/ProfileTab.tsx
```diff
@@ -1,16 +1,123 @@
...
+    handleNonInputValueChange,
+    data,
+    errors
+  } = useForm<EditProfileForm>({
+    initialValues,
+    validations
+  })
+
+  const isPhotoChanged = photo === null || Boolean(photo.src)
+  const isDataChanged = isDirty || isPhotoChanged
```
byavdoshniak, line 55: can we somehow make useForm handle photo as well?

src/containers/edit-profile/profile-tab/ProfileTab.tsx
```diff
@@ -1,16 +1,123 @@
...
+
+  const blocker = useBlocker(isDataChanged)
```
byavdoshniak, line 57: Nice one

src/containers/edit-profile/profile-tab/ProfileTab.tsx
```diff
@@ -1,16 +1,123 @@
...
+    handleBlur,
+    handleNonInputValueChange,
+    data,
+    errors
+  } = useForm<EditProfileForm>({
+    initialValues,
+    validations
+  })
+
+  const isPhotoChanged = photo === null || Boolean(photo.src)
```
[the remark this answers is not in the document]
iva-stasia: When receiving a nullish value for a user's photo, we set the default value `{src: '', name: ''}` for the `photo` variable to make it easier to handle photo changes. This way, this condition won't be true if a user doesn't have a photo to initially

src/containers/edit-profile/profile-tab/ProfileTab.tsx
iva-stasia: Delegated the handling of photo changes to the `useForm` hook. It really helped to make the code more organized and clean, thanks.

Issue #1819, opened 2024-04-22
(SP:) [Tutor cooperarion] Implement tutor view on saved cooperation


asya982:
[Link to Figma](https://www.figma.com/file/liJZDYhUnDP15Pi9bipDcv/Space2Study-(Students)?type=design&node-id=13019-49302&mode=design&t=B67KY15vrrgPzRq4-0)

PR #1846, opened 2024-04-26
Implement lesson view for students

iva-stasia:
Implemented lesson view for students.

**To try out these changes**, you need to pull changes from the [backend PR](https://github.com/ita-social-projects/SpaceToStudy-BackEnd/pull/775).
Since we don't have cooperation activities view yet (is currently being implemented #1819), you need to write the path manually, for example `http://localhost:3000/my-cooperations/6628deaed9799133050a9c7a/lesson-details/6628e141d9799133050a9d08`.

- `FileEditor.tsx`: The Rubik font is added and set as the default (since it's the font used in our app).
- `AppBreadCrumbs.tsx`, `authRoutes.ts`, `authRouter.tsx`:  added a path and changed the breadcrumbs so that users can navigate correctly using breadcrumbs.
- `LessonDetails.tsx`: added a condition that indicates whether the lesson is editable; changed styles to match the mockup.
- `LessonDetails.spec.jsx`: added tests to achieve 100% coverage.
- Added the `DOMPurify` library to make setting HTML safe.

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/110097862/307e1de8-97e4-4220-9112-67239f0d8177

Screenshot 2024-04-25 at 23 04 28


src/components/file-editor/FileEditor.tsx
```diff
@@ -19,7 +19,9 @@ const FileEditor: FC<FileEditorProps> = ({ onEdit, value }) => {
         toolbar:
           'undo redo | blocks fontsize | bold italic underline strikethrough | ltr rtl | link image media table mergetags | align lineheight | tinycomments | checklist numlist bullist indent outdent accordion | removeformat',
         content_style:
-          'body { font-family:Helvetica,Arial,sans-serif; font-size:14px }'
+          '@import url("https://fonts.googleapis.com/css2?family=Rubik:wght@400;600;800&display=swap"); body { font-family:Rubik; font-size:14px }',
+        font_family_formats:
+          'Andale Mono=andale mono,times; Arial=arial,helvetica,sans-serif; Arial Black=arial black,avant garde; Book Antiqua=book antiqua,palatino; Comic Sans MS=comic sans ms,sans-serif; Courier New=courier new,courier; Georgia=georgia,palatino; Helvetica=helvetica; Impact=impact,chicago; Symbol=symbol; Tahoma=tahoma,arial,helvetica,sans-serif; Terminal=terminal,monaco; Times New Roman=times new roman,times; Trebuchet MS=trebuchet ms,geneva; Verdana=verdana,geneva; Webdings=webdings; Wingdings=wingdings,zapf dingbats; Rubik=rubik'
```
mhevyk, line 24: Seems like there is a list of key-value pairs with font name in editor and it`s actual value, separated by semicolons. Would it be better if we move key value pairs into separate object and use joining on it like following:

```jsx
const availableEditorFonts = {
     "Andale Mono": "andale mono"
}

font_family_formats: Object.entries(availableEditorFonts).map(entry => entry.join("=")).join("; ")
```

What do you think about this approach? Maybe adding or removing fonts will be more convenient using object

tests/unit/pages/lesson-details/LessonDetails.spec.jsx
```diff
@@ -65,13 +70,25 @@ describe('LessonDetails', () => {
...
+    expect(attachment).toBeVisible()
+
+    fireEvent.click(title)
+
+    await waitFor(() => {
+      expect(attachment).not.toBeVisible()
+    })
+  })
+
+  it('should ...', async () => {
```
mhevyk, line 89: Why description of test is `should ...`?

src/pages/lesson-details/LessonDetails.tsx
```diff
@@ -76,34 +78,45 @@ const LessonDetails = () => {
   const items = [
     {
       title: 'lesson.content',
-      content: <div dangerouslySetInnerHTML={{ __html: response.content }} />
+      content: (
+        <Box
+          dangerouslySetInnerHTML={{ __html: response.content }}
```
mhevyk, line 83: Using `dangerouslySetInnerHTML` is dangerous in scope of XSS attacks for example. Maybe we need to use a package for it? For example `dompurify` is a good option used in many cases

Maybe would be useful: https://deadsimplechat.com/blog/how-to-safely-use-dangerouslysetinnerhtml-in-react/

src/components/file-editor/FileEditor.tsx
iva-stasia: Thanks for the suggestion 👍  I decided to move all the configurations to the constants file. Hopefully, it will be easier to maintain.

tests/unit/pages/lesson-details/LessonDetails.spec.jsx
iva-stasia: Thanks for pointing this out. Forgot to apply changes after stashing.

OliaHrytsak (approved):

GeorgeShvab (approved):

D1nviz (approved):

src/pages/lesson-details/LessonDetails.tsx
iva-stasia: Thanks for the suggestion, I used the `DOMPurify` library and now it looks safer indeed.

mhevyk (approved):

byavdoshniak (approved):

PR #1853, opened 2024-05-01
deleted translation + added scss/lin to gitignore

asya982:
I decided to delete tranlsation in enroll offer - because in all site we have no translations for both categories and level 
So if needed we can create another issue to integrate translations for this fields, which will require refactoring on back-end side as well
I guess it's not our priority for now, so let's all will be in English as proper names

Also i added `scss/lib` folder, where our complied css stored in `.gitignore` file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,7 @@ test-report.xml
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+
+# scss builds
+/src/scss/lib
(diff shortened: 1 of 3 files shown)


D1nviz (approved):

Mav-Ivan (approved):

GeorgeShvab (approved):

mhevyk (approved):

OliaHrytsak (approved):

.gitignore
```diff
@@ -27,3 +27,7 @@ test-report.xml
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+
+# scss builds
+/src/scss/lib
```
iva-stasia, line 33: Seems like a newline is missing here

iva-stasia (approved):

PR #1856, opened 2024-05-02
Replaced react-beatiful-dnd with react-dnd-kit

D1nviz:
### So I decided to use react-dnd-kit library because: 
- Zero dependent
- Has many features (almost similar to react-dnd)
- Super modern and easy to read docs

Also, it gives us cool animations, that make user experience better 
### Showcase:
https://github.com/ita-social-projects/SpaceToStudy-Client/assets/118522791/38ec5c10-9e39-48df-b7bc-d2b0e0e71fa0

I will create an issue for covarage to this components


src/containers/course-section/resources-list/ResourcesList.tsx
```diff
@@ -77,47 +72,71 @@ const ResourcesList: FC<ResourcesListProps> = ({
...
+    },
+    [setResources]
+  )
+
+  const renderItem = (item: CourseResources, isDragOver = false) => (
+    <SortableWrapper
+      id={item._id}
+      key={item._id}
+      onDragEndStyles={styles.section(isDragOver)}
+      onDragStartStyles={styles.section(true)}
```
mhevyk, line 101: What does `true` in `styles.section(true)` mean?

src/containers/course-section/resources-list/ResourcesList.tsx
```diff
@@ -77,47 +72,71 @@ const ResourcesList: FC<ResourcesListProps> = ({
...
-              </Box>
-            )}
-          </Droppable>
-        )}
-      </DragDropContext>
-    </Box>
+    <DndContext
+      onDragCancel={() => {
+        setActive(null)
+      }}
```
mhevyk, line 119: I suggest moving `onDragCancel` to `handleDragCancel` function above return

src/containers/course-section/resources-list/ResourcesList.tsx
```diff
@@ -77,47 +72,71 @@ const ResourcesList: FC<ResourcesListProps> = ({
...
+      onDragEnd={handleDragEnd}
+      onDragStart={({ active }) => {
```
mhevyk, line 121: Same here

src/containers/course-section/resources-list/ResourcesList.tsx
```diff
@@ -77,47 +72,71 @@ const ResourcesList: FC<ResourcesListProps> = ({
...
+            items={items.map((item) => item._id)}
+            strategy={verticalListSortingStrategy}
+          >
+            <Box sx={styles.root}>{resourceItems}</Box>
+          </SortableContext>
+          <DragOverlay>
+            {activeItem && renderItem(activeItem, true)}
+          </DragOverlay>
+        </>
+      )}
```
mhevyk, line 138: Please move it outside `return` statement to avoid polluting jsx

src/containers/course-sections-list/CourseSectionsList.tsx
```diff
@@ -45,7 +50,30 @@ const CourseSectionsList: FC<CourseSectionsListProps> = ({
...
+  const onDragEnd = useCallback(
+    (event: DragEndEvent) => {
+      const { active, over } = event
+
+      if (active && over) {
+        const activeIndex = items.findIndex((item) => item.id === active.id)
+        const overIndex = items.findIndex((item) => item.id === over.id)
+        if (activeIndex !== -1 && overIndex !== -1) {
+          setSectionsItems(arrayMove(items, activeIndex, overIndex))
+        }
```
mhevyk, line 70: I have seen this code before. Maybe we can create custom hook something like `useDndSensor` and extract repeated logic (`useSensors`, `onDrag` etc)

src/containers/course-sections-list/CourseSectionsList.tsx
```diff
@@ -121,80 +123,79 @@ const CourseSectionsList: FC<CourseSectionsListProps> = ({
...
+                transformOrigin: {
+                  vertical: 'top',
+                  horizontal: 'center'
+                },
+                anchorOrigin: {
+                  vertical: 'bottom',
+                  horizontal: 'center'
+                },
+                sx: styles.menuRoot
+              })}
```
mhevyk, line 155: Move it before `return` statement

src/containers/course-sections-list/CourseSectionsList.tsx
```diff
@@ -121,80 +123,79 @@ const CourseSectionsList: FC<CourseSectionsListProps> = ({
...
-                {provided.placeholder}
-              </Box>
-            )}
-          </Droppable>
-        )}
-      </DragDropContext>
-    </Box>
+    <DndContext
+      onDragCancel={() => {
+        setActive(null)
```
mhevyk, line 180: Same here

src/containers/questions-list/QuestionsList.tsx
```diff
@@ -24,76 +38,82 @@ interface QuestionsListProps {
...
+          onCancel={onEditCancel}
+          question={item}
+          setQuestions={setItems}
+        />
+      ) : (
+        <Question
+          question={item}
+          setEditableItemId={setEditableItemId}
+          setQuestions={setItems}
+        />
```
mhevyk, line 92: Also move out of jsx

src/containers/sortable-wrapper/SortableWrapper.tsx
```diff
@@ -0,0 +1,43 @@
+import { CSSProperties, FC } from 'react'
+import { useSortable } from '@dnd-kit/sortable'
+import { CSS } from '@dnd-kit/utilities'
+import { Box, SxProps } from '@mui/material'
```
mhevyk, line 4: Use default import of `Box` instead of named

src/containers/course-section/resources-list/ResourcesList.tsx
D1nviz: It's mean, that if state `true`, this piece of styles will be added
```
section: (isDragging: boolean) => ({
    mb: '20px',
    backgroundColor: 'basic.white',
    borderRadius: '6px',
    ...(isDragging && {
      boxShadow: `0px 3px 16px 2px ${palette.primary[300]}`,
      border: `2px solid ${palette.primary[300]}`
    })
  }),
```

src/containers/course-section/resources-list/ResourcesList.tsx
D1nviz: I agree, fixed. I Mooved it to a custom hook `useDndSensor`

src/containers/course-section/resources-list/ResourcesList.tsx
D1nviz: changed

src/containers/course-section/resources-list/ResourcesList.tsx
D1nviz: thank you for notice, fixed

src/containers/course-sections-list/CourseSectionsList.tsx
D1nviz: Yes, it's good idea. Also we have a problem with different id key, so i created generic and now it's works

src/containers/sortable-wrapper/SortableWrapper.tsx
D1nviz: didn't notice, thank you. Fixed

src/containers/questions-list/QuestionsList.tsx
D1nviz: Changed

src/containers/course-section/resources-list/ResourcesList.tsx
mhevyk: Makes sense, thanks

src/containers/course-sections-list/CourseSectionsList.tsx
mhevyk: Good for you!

src/hooks/use-dnd-sensor.ts
```diff
@@ -0,0 +1,59 @@
...
+  Active
+} from '@dnd-kit/core'
+import { arrayMove } from '@dnd-kit/sortable'
+
+type IdProp = 'id' | '_id'
+
+const useDndSensor = <T extends { [K in IdProp]?: string }>(
+  items: T[],
+  setItems: (items: T[]) => void,
+  idProp: IdProp
```
mhevyk, line 18: Maybe it would be better interface for hook if we use object instead of three different parameters? I mean usage like following:

```jsx
const {...} = useDndSensor({
    items: ...,
    setItems: ...,
    idProp: ...
})
```

What do you think about it?

src/context/sortable-context.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const SortableItemContext = createContext<SortableItemProvideContext>({
+  attributes: {
+    role: 'button',
+    tabIndex: 0,
+    'aria-disabled': false,
+    'aria-pressed': false,
+    'aria-roledescription': '',
+    'aria-describedby': ''
```
mhevyk, line 23: Can you please explain what do these attributes do?

src/containers/course-sections-list/CourseSectionsList.styles.ts
```diff
@@ -26,7 +26,8 @@ export const styles = {
     fontSize: '30px',
     transform: 'rotate(90deg)',
     color: 'primary.400',
-    cursor: 'pointer'
+    cursor: 'grab',
+    touchAction: 'none'
```
mhevyk, line 30: I have noticed that these styles are quite common inside drag icons. Maybe we can adjust `DragHandle.tsx` and add these styles only in one place to all handlers or just create some common style to reuse it? What do you think about it?

src/context/sortable-context.tsx
D1nviz: It's required for  interface `DraggableAttributes` and without it drag&drop will not work

src/hooks/use-dnd-sensor.ts
D1nviz: Thank you, it's cool idea! Refactored

src/components/drag-handle/DragHandle.tsx
```diff
@@ -0,0 +1,21 @@
+import { FC } from 'react'
+import { Box, SxProps } from '@mui/material'
```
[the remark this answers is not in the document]
D1nviz: fixed

src/containers/course-sections-list/CourseSectionsList.tsx
```diff
@@ -1,32 +1,25 @@
...
+import { DndContext, DragOverlay } from '@dnd-kit/core'
+import { SortableContext, verticalListSortingStrategy } from '@dnd-kit/sortable'
+import { useTranslation } from 'react-i18next'
 import Box from '@mui/material/Box'
-import DragIndicatorIcon from '@mui/icons-material/DragIndicator'
-import Divider from '@mui/material/Divider'
-import MenuItem from '@mui/material/MenuItem'
-import Typography from '@mui/material/Typography'
 import Crop75Icon from '@mui/icons-material/Crop75'
+import { Divider, MenuItem, Typography } from '@mui/material'
```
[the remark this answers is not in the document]
D1nviz: fixed

src/containers/course-sections-list/CourseSectionsList.styles.ts
D1nviz: i created a separate file for dragHandle style. Fixed

src/hooks/use-dnd-sensor.ts
```diff
@@ -0,0 +1,59 @@
...
+  TouchSensor,
+  DragStartEvent,
+  DragEndEvent,
+  Active
+} from '@dnd-kit/core'
+import { arrayMove } from '@dnd-kit/sortable'
+
+type IdProp = 'id' | '_id'
+
+const useDndSensor = <T extends { [K in IdProp]?: string }>(
```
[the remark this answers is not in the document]
D1nviz: I used this pattern. What do you think about it?

t```
ype IdType1 = { id: string | number; _id?: string | number }
type IdType2 = { id?: string | number; _id: string | number }

type IdType = IdType1 | IdType2

interface UseDndSensorProps<T extends IdType> {
  items: T[]
  setItems: (items: T[]) => void
  idProp: 'id' | '_id'
}

const useDndSensor = <T extends IdType>({
```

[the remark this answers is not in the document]
Comment (mhevyk) [reply src/hooks/use-dnd-sensor.ts]: I think it is a pretty good solution here because as you had shown currently we have items, that for some reason pass down both `_id` and `id` prop, which most likely incorrect, but this is not related to this issue.

I also suggest renaming from `IdType` to `ObjectWithId` to be more precise in naming. Then there is no need to create types `IdType1` and `IdType2`, just use them inline inside `ObjectWithId`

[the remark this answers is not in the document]
Comment (D1nviz) [reply src/hooks/use-dnd-sensor.ts]: Changed

File: src/containers/course-section/resources-list/ResourcesList.styles.ts [hunk 16]
```diff
@@ -9,6 +9,7 @@ export const styles = {
     position: 'relative',
     backgroundColor: 'basic.white',
     borderRadius: '6px',
+    zIndex: isDragging ? 10 : 1,
```
asya982, line 12: Default value for `z-index` is 0, if we can drag only 1 item per time, maybe it's better to set up it like this, should still work, and we don't overload page with redundant high z-index
What do you think?
```suggestion
    zIndex: isDragging ? 1 : 0,
```

mhevyk (approved):

asya982 (approved):

GeorgeShvab (approved):

byavdoshniak (approved):

PR #1857, opened 2024-05-04
Hotfix/tooltip for disable button_create or edit quiz container_fix bug

OliaHrytsak:
After edit tooltip to the button 'Create new question' it doesn't work correct, because MUI component Tooltip, doesn't pass ref to the castom components like AppButton.
Because of that, was added <span> wrapper to AppButton component that helps to show disable button with tooltip without errors
From ofаicial MUI documentation
image
Link
https://mui.com/material-ui/react-tooltip/
How it works now

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/127191734/6d4a57e8-b50b-47ff-8e20-29e2f5190ea6

Also I've add corsor: 'pointer' to the User Avatar in this hotfix PR as I forgot about it before(

### DIFF
--- a/src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.styles.ts
+++ b/src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.styles.ts
@@ -61,6 +61,9 @@ export const styles = {
     '& button': {
       gap: '12px',
       width: '100%'
+    },
+    '& span': {
+      width: '100%'
     }
   },
 
--- a/src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
+++ b/src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
@@ -204,7 +204,6 @@ const CreateOrEditQuizContainer = ({
   const CreateQuestionButton = (
     <Tooltip
       arrow
-      disableTouchListener
       placement={PositionEnum.Top}
       title={
         isCreationOpen ? t('myResourcesPage.quizzes.savePreviousQuestion') : ''
@@ -212,6 +211,7 @@ const CreateOrEditQuizContainer = ({
     >
       <span>
         <AppButton
+          disabled={isCreationOpen}
           onClick={onOpenCreateQuestion}
           size={SizeEnum.ExtraLarge}
           variant={ButtonVariantEnum.Tonal}
--- a/src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
+++ b/src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
@@ -204,19 +204,22 @@ const CreateOrEditQuizContainer = ({
   const CreateQuestionButton = (
     <Tooltip
       arrow
+      disableTouchListener
       placement={PositionEnum.Top}
       title={
         isCreationOpen ? t('myResourcesPage.quizzes.savePreviousQuestion') : ''
       }
     >
-      <AppButton
-        onClick={onOpenCreateQuestion}
-        size={SizeEnum.ExtraLarge}
-        variant={ButtonVariantEnum.Tonal}
-      >
-        {t('myResourcesPage.quizzes.createNewQuestion')}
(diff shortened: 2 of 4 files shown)


D1nviz (approved):

src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
```diff
@@ -209,14 +209,17 @@ const CreateOrEditQuizContainer = ({
...
     >
-      <AppButton
-        onClick={onOpenCreateQuestion}
-        size={SizeEnum.ExtraLarge}
-        variant={ButtonVariantEnum.Tonal}
-      >
-        {t('myResourcesPage.quizzes.createNewQuestion')}
-        <EditIcon fontSize={SizeEnum.Small} />
-      </AppButton>
+      <span>
```
mhevyk, line 212: Maybe we should finally refactor AppButton to accept a ref? 
image

As you can see from mui docs, simple Button works here. We can avoid dom elements nesting with such approach

src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.styles.ts
```diff
@@ -61,6 +61,9 @@ export const styles = {
     '& button': {
       gap: '12px',
       width: '100%'
+    },
+    '& span': {
+      width: '100%'
```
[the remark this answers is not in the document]
OliaHrytsak: In this case span and button in one Box with styles.functionalButtons
As I wraped AppButton into <span> I've to fix width of this <span> as a childe of Box 
If I dont't wrap AbbButton into <span>, ref from button doesn't pass to Tooltip and styles of this button are broken because of tooltip element

I don't know how explaine it in another way, but it also an example froь official MUI documentation for using toltip with disable button

src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
OliaHrytsak: I did such an experiment) After adding ref and forwardRef to AppButton without <span> error disappeared but I can't add disabled={isCreationOpen} and it is not wery well user experience.... And also variant={ButtonVariantEnum.Tonal} doesn't work((((


https://github.com/ita-social-projects/SpaceToStudy-Client/assets/127191734/fbb20d30-cbaa-4f29-b1f3-b4c30153d6ba

asya982 (approved):

iva-stasia (approved):

GeorgeShvab (approved):

src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
mhevyk: I think there might be another problem related to it because ref should not impact on styles

src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.styles.ts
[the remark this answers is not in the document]
mhevyk: If a better solution can't be found, we can keep with your one

src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
OliaHrytsak: I've understood why styles is broken after adding tooltip to the AppButton !
The main reason isn't forwardRef, of course)
This button, after wrapping in Tooltip, has property data-mui-internal-clone-element='true': This attribute indicates that the element is being cloned internally by Material-UI. Material-UI sometimes clones elements for various reasons such as event handling, state management, or accessibility enhancements. It's used internally by the library and should generally not be manipulated directly. (our s2s_button styles obviously don't work with cloned elements)

image

And when we add <span> it prevents manipulating of the nested button elemetn by Tooltip
image

In that case <span> works as a element that allows as add Tooltip to castom element without manipulating of this element

mhevyk (approved):

PR #1873, opened 2024-05-07
Removed cooperation context and used redux toolkit instead

mhevyk:
Cooperation context was removed, I added new slice with data from that context. Also I fixed tests


src/containers/my-cooperations/cooperation-activities-list/CooperationActivitiesList.tsx
```diff
@@ -17,7 +17,7 @@ const CooperationActivitiesList = ({
   handleNonInputValueChange
 }: CooperationActivitiesListProps) => {
   const { selectedCourse, isAddedClicked, currentSectionIndex } =
-    useCooperationContext()
+    useAppSelector((state) => state.cooperations)
```
GeorgeShvab, line 20: Maybe we can create selector callback function in `cooperationsSlice.ts` and export it? This way we won't need to repeat the same `(store) => store.cooperations` in different places (for now repeated in `CourseSectionsList.tsx`, `CooperationActivitiesList.tsx` and `CooperationDetails.tsx`). What do you think?

mhevyk: I agree with you. Created `cooperationsSelector` and used it like following:

```jsx
const {...} = useAppSelector(cooperationsSelector);
```

src/constants/index.ts
```diff
@@ -27,3 +27,7 @@ export const itemsLoadLimit = {
   mobile: 6,
   default: 12
 }
+
+export const sliceNames = {
+  cooperations: 'cooperationsSlice'
```
asya982, line 32: Maybe this object should be moved to `/redux` separate file like `constants.ts`, since it's only will be used in redux slices? What do you think?

mhevyk: Created `redux.constants.ts` and moved object there. Thanks for suggestion

byavdoshniak (approved):

OliaHrytsak (approved):

asya982 (approved):

D1nviz (approved):

GeorgeShvab (approved):

PR #1887, opened 2024-05-11
Replaced snackbar context with redux toolkit

GeorgeShvab:
- Created snackbar slice added it to the redux store
- Created `AppSnackbar` component
- Replaced `useSnackbarContext` and `setAlert` with `useDispatch` and `dispatch`+`openAlert`
- Changed tests accordingly. Previously, snackbar context wrapped children in `renderWithProviders` (In `PopupsProvider`). But with redux, Snackbar can't be added on `renderWithProviders` level because many tests use it and some of them mocks `react-redux` import (`SignupDialog` for example). So I added `TestSnackbar` which adds Snackbar to tested components.


src/containers/layout/app-snackbar/AppSnackbar.tsx
```diff
@@ -0,0 +1,36 @@
+import { useAppDispatch, useAppSelector } from '~/hooks/use-redux'
+import Snackbar from '@mui/material/Snackbar'
+import Alert from '@mui/material/Alert'
+import { Box } from '@mui/material'
```
mhevyk, line 4: Use default omport here

src/containers/add-resources/AddResources.tsx
```diff
@@ -60,12 +61,14 @@ const AddResources = <T extends CourseResources | Question>({
...
   const onResponseError = useCallback(
     (error: ErrorResponse) => {
-      setAlert({
-        severity: snackbarVariants.error,
-        message: error ? `errors.${error.code}` : ''
-      })
+      dispatch(
+        openAlert({
+          severity: snackbarVariants.error,
+          message: error ? `errors.${error.code}` : ''
```
asya982, line 67: Maybe it's better to set up some default error instead of just empty string? I saw below we have UNKNOWN_ERROR constant
I saw some alert with just icon and it'll look a little bit confusing, how do you think?
And if you agree, please don't forget to change it in all occurrences🥹

src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx
```diff
@@ -80,24 +81,28 @@ const CreateOrEditQuizContainer = ({
...
-        ? t(`errors.${error.code}`, {
-            message: getErrorMessage(error.message)
-          })
-        : ''
-    })
+    dispatch(
+      openAlert({
+        severity: snackbarVariants.error,
+        message: error
+          ? t(`errors.${error.code}`, {
```
asya982, line 100: Isn't here will be some kind of error, since you already included `t` in snack bar component?

OliaHrytsak (approved):

OliaHrytsak:
In few tests there was - import { vi } from 'vitest' - it wasn't add by you, but I think it would be good idea to delete this unnecessary imports in this pull request.

GeorgeShvab:
> In few tests there was - import { vi } from 'vitest' - it wasn't add by you, but I think it would be good idea to delete this unnecessary imports in this pull request.

Thanks for noticing, I removed unnecessary imports.

src/containers/layout/app-snackbar/AppSnackbar.tsx
GeorgeShvab: Thanks, it was replaced with default import.

src/containers/add-resources/AddResources.tsx
GeorgeShvab: I agree with you. I made function like this ```
const getErrorKey = (error) =>
  `errors.${error && error.code ? error.code : 'UNKNOWN_ERROR'}`
```
 to return error translation key. Used it in all the occurrences.

File: src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx (hunk 3 above)
Comment (GeorgeShvab) [reply src/containers/my-quizzes/create-or-edit-quiz-container/CreateOrEditQuizContainer.tsx]: Thanks for noticing. There won't be error actually. In other places, I removed `t` from `openAlert` call, but here we also pass some options, not just string, so I kept it before. But it really looks unnecessary, to use `t` here and inside of `AppSnackbar`, so I added possibility to pass translation options directly to snackbar redux state. So now snackbar message type looks like this 
```
interface ExtendedSnackbarMessage {
  text: string
  options: WritableDraft<TOptions>
}
type SnackbarMessage = string | ExtendedSnackbarMessage
```
`WritableDraft` is used because some of TOptions properties use `readonly`, so ts throws error without it. Besides, we don't change those properties directly.

asya982: Nice solution!

asya982 (approved):

D1nviz (approved):

byavdoshniak (approved):

PR #1893, opened 2024-05-14
Added tests for professional tab after ui changes

mhevyk:
Wrote tests to Improve coverage as much as possible for `professional-info-tab`

image

When speaking about `AddProfessionalCategoryModal.tsx`, following lines could not be tested in any way, so I kept it as it is because coverage is decent enough 

image

image

As for `ProfessionalCategory.tsx`, uncoreved lines contain `TODO`s, so i can't test them now. These lines should be tested when actual logic with backend is added:

image

image

### DIFF
--- a/tests/test-utils.jsx
+++ b/tests/test-utils.jsx
@@ -8,7 +8,6 @@ import { theme } from '~/styles/app-theme/custom-mui.styles'
 import PopupsProvider from '~/PopupsProvider'
 import cooperationsReducer from '~/redux/features/cooperationsSlice'
 
-import { vi } from 'vitest'
 import MockAdapter from 'axios-mock-adapter'
 import { axiosClient } from '~/plugins/axiosClient'
 
--- a/tests/test-utils.jsx
+++ b/tests/test-utils.jsx
@@ -3,7 +3,7 @@ import { Provider } from 'react-redux'
 import { configureStore } from '@reduxjs/toolkit'
 import reducer from '~/redux/reducer'
 import { StyledEngineProvider, ThemeProvider } from '@mui/material/styles'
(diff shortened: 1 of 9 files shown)


tests/test-utils.jsx
```diff
@@ -3,7 +3,7 @@ import { Provider } from 'react-redux'
 import { configureStore } from '@reduxjs/toolkit'
 import reducer from '~/redux/reducer'
 import { StyledEngineProvider, ThemeProvider } from '@mui/material/styles'
-import { render, waitFor } from '@testing-library/react'
+import { screen, render, waitFor, fireEvent, act } from '@testing-library/react'
```
OliaHrytsak, line 6: You can remove 'render' from import.

tests/test-utils.jsx
```diff
@@ -49,3 +49,17 @@ export const waitForTimeout = (callback, options) => {
...
+    fireEvent.change(selectLike, { target: { value: option } })
+  })
+
+  const selectedOption = screen.getByText(option)
+  await act(async () => {
+    fireEvent.click(selectedOption)
+  })
+
+  expect(selectLike.value).toBe(option)
+}
```
OliaHrytsak, line 65: Good idea to move repeating function into test-utils!

OliaHrytsak (approved):

tests/test-utils.jsx
mhevyk: I can't remove it because it fails. But I found `vi` import and I removed it instead

asya982 (approved):

GeorgeShvab (approved):

D1nviz (approved):

Mav-Ivan (approved):

PR #1900, opened 2024-05-16
Fixed page scroll on new lesson page

GeorgeShvab:
- Basically, I only added `ui_mode: 'split'` to init options. All the other changes are refactoring, I moved static init options to constants file and made a function to add other dynamic options.

- Library we use for text redactor recalculated menu position on document `scroll` event. I guess the problem was that in our application such event was not dispatched as scroll occurred not in `body` but in `main` element (even scrollbar indicates that scrollable part is not entire page). Developers of the library recommend to use `ui_mode: 'split'` when redactor is in scrollable element. 

- As you can see, menus appear below header, I don't think it can be changed with this library and with scroll fix.

Before and after changes demonstrations are below.

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/62070431/42154d81-d779-426b-814d-aca90dc3a5ff

https://github.com/ita-social-projects/SpaceToStudy-Client/assets/62070431/90bd8907-4095-4ca3-83dd-14d8b5394d21

### DIFF
--- a/src/components/file-editor/FileEditor.tsx
+++ b/src/components/file-editor/FileEditor.tsx
@@ -1,6 +1,6 @@
 import { FC } from 'react'
 import { Editor } from '@tinymce/tinymce-react'
-import { fileEditorConfig } from './FileEditor.constants'
+import { getEditorInitOptions } from './FileEditor.constants'
 import { useTranslation } from 'react-i18next'
 
 interface FileEditorProps {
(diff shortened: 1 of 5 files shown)


Mav-Ivan:
Don`t forget to improve the coverage

Mav-Ivan (approved):

src/components/file-editor/FileEditor.tsx
```diff
@@ -1,6 +1,6 @@
 import { FC } from 'react'
 import { Editor } from '@tinymce/tinymce-react'
-import { fileEditorConfig } from './FileEditor.constants'
+import { getEditorInitOptions } from './FileEditor.constants'
```
D1nviz, line 3: Use alias import here

byavd (approved):

asya982 (approved):

OliaHrytsak (approved):

src/components/file-editor/FileEditor.tsx
GeorgeShvab: Thanks for noticing. Used alias here.

D1nviz (approved):

PR #1918, opened 2024-05-26
Synchronized structure of availability object in cooperations and fixed small bugs

mhevyk:
The main task here was to fix availability object in cooperation activities according to server's response.

Also I included following changes:
1. Error regarding to `key` in `CourseSectionsList.tsx`
2. Typescript issues (renaming `CourseResources` to `CourseResource`, renaming `openModalFunc` to `OpenModalFunction`)
3. Used more intuitive variable names and reused code as much as possible

Also I performed some investigation for #1919. We should ask our experts about this task, most likely we will need to rewrite a lot of cooperation logic because current implementation is not suitable for moving data efficiently to the top-level components

### DIFF
--- a/src/containers/course-section/resource-item/ResourceItem.tsx
+++ b/src/containers/course-section/resource-item/ResourceItem.tsx
@@ -1,4 +1,6 @@
-import { FC, ReactElement, useCallback, useEffect } from 'react'
+import { FC, useCallback, useEffect } from 'react'
+import { useTranslation } from 'react-i18next'
+
 import Box from '@mui/material/Box'
 import IconButton from '@mui/material/IconButton'
 import CloseIcon from '@mui/icons-material/Close'
@@ -7,20 +9,21 @@ import TextField from '@mui/material/TextField'
 import { AdapterDateFns } from '@mui/x-date-pickers/AdapterDateFns'
 import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider'
 import EditIcon from '@mui/icons-material/Edit'
-import { useTranslation } from 'react-i18next'
 
 import AppSelect from '~/components/app-select/AppSelect'
 import IconExtensionWithTitle from '~/components/icon-extension-with-title/IconExtensionWithTitle'
 import { useResourceAvailabilityContext } from '~/context/resources-availability-context'
+
 import openIcon from '~/assets/img/cooperation-details/resource-availability/open-icon.svg'
 import openFrom from '~/assets/img/cooperation-details/resource-availability/open-from.svg'
 import closedIcon from '~/assets/img/cooperation-details/resource-availability/closed-icon.svg'
 
 import {
-  CourseResources,
+  CourseResource,
   ResourceAvailabilityStatusEnum,
   ResourcesAvailabilityEnum,
   ResourcesTabsEnum as ResourcesTypes,
+  SetResourseAvailability,
   SizeEnum
 } from '~/types'
 
@@ -29,27 +32,19 @@ import { resourcesData } from '~/containers/course-section/CourseSectionContaine
 import { styles } from '~/containers/course-section/resource-item/ResourceItem.styles'
 
 interface ResourceItemProps {
-  resource: CourseResources
-  deleteResource?: (resource: CourseResources) => void
-  setResourceAvailability?: (
-    id: string,
-    availability: ResourceAvailabilityStatusEnum,
-    openFromDate: string | null
-  ) => void
-  editResource?: (resource: CourseResources) => void
+  resource: CourseResource
+  deleteResource?: (resource: CourseResource) => void
+  setResourceAvailability?: SetResourseAvailability
+  editResource?: (resource: CourseResource) => void
   isView?: boolean
 }
 
-interface AvailabilityIconsProps {
-  open: string
-  openFrom: string
-  closed: string
-}
+const { Open, OpenFrom, Closed } = ResourceAvailabilityStatusEnum
 
-const availabilityIcons: AvailabilityIconsProps = {
-  open: openIcon,
-  openFrom: openFrom,
-  closed: closedIcon
+const availabilityIcons: Record<ResourceAvailabilityStatusEnum, string> = {
+  [Open]: openIcon,
+  [OpenFrom]: openFrom,
+  [Closed]: closedIcon
 }
 
 const ResourceItem: FC<ResourceItemProps> = ({
@@ -59,20 +54,20 @@ const ResourceItem: FC<ResourceItemProps> = ({
   editResource,
   isView = false
 }) => {
-  const onDeleteResource = () => {
+  const handleDeleteResource = () => {
     deleteResource?.(resource)
   }
 
-  const onEditResource = () => {
+  const handleEditResource = () => {
     editResource?.(resource)
   }
 
   const { t } = useTranslation()
 
-  const { resourceAvailability, isCooperation } =
+  const { resourceAvailability: allResourcesAvailability, isCooperation } =
     useResourceAvailabilityContext()
 
-  const setResourceIcon = (): ReactElement | undefined => {
+  const renderResourceIcon = () => {
     if (resource.resourceType === ResourcesTypes.Lessons) {
       return resourcesData.lessons.icon
     } else if (resource.resourceType === ResourcesTypes.Quizzes) {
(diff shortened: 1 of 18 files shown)


D1nviz (approved):

D1nviz (approved):

src/containers/course-sections-list/CourseSectionsList.tsx
```diff
@@ -146,11 +146,10 @@ const CourseSectionsList: FC<CourseSectionsListProps> = ({
     )
 
     return (
-      <>
+      <Fragment key={item.id}>
```
dmtrth25, line 149: don't understood. 
Why did you replace `<>` with just Fragment instead and used `key` value. Are you use some method for `CourseSectionsList` ?

src/containers/course-section/resource-item/ResourceItem.tsx
```diff
@@ -59,109 +54,102 @@ const ResourceItem: FC<ResourceItemProps> = ({
...
     useResourceAvailabilityContext()
 
-  const setResourceIcon = (): ReactElement | undefined => {
+  const renderResourceIcon = () => {
     if (resource.resourceType === ResourcesTypes.Lessons) {
       return resourcesData.lessons.icon
     } else if (resource.resourceType === ResourcesTypes.Quizzes) {
       return resourcesData.quizzes.icon
     }
   }
```
dmtrth25, line 76: I guess it can be improved.
What do you think ?

src/containers/course-sections-list/CourseSectionsList.tsx
mhevyk: Thanks for question.  It was a tough one to fix and i struggled with it a lot. The reason I have added it is that `sectionItem` function is used inside of `map` in following code snippet:

```jsx
const sectionItems =
    items.length === 0
      ? clearCoorperationMenu
      : items.map((item) => sectionsItem(item))
```

When I was investigating error in console, it said that the error is present inside `<SortableContext items={items} strategy={verticalListSortingStrategy}>` in the following code:

```jsx
const courseSectionContent = enabled && (
    <>
      <SortableContext items={items} strategy={verticalListSortingStrategy}>
        {sectionItems}
      </SortableContext>
      <DragOverlay>{activeItem && sectionsItem(activeItem, true)}</DragOverlay>
    </>
  )
```

That's confused me, because `SortableContext` is nor being used inside any kind of loop. So I came up with solution that you can see now, because `cooperationMenu` and `SortableWrapper` are repeated on the same level and we should provide a wrapper for it to make an error disappear. I decided not no use extra `Box`, just sticked to `React`'s `Fragment`

I hope I have answered your question and it because clearer for you

src/containers/course-section/resource-item/ResourceItem.tsx
mhevyk: I found this solution rather good for this use case. I also considered following solution:

```jsx
let icon;

if (resource.resourceType === ResourcesTypes.Lessons) {
     icon = resourcesData.lessons.icon
}

if (resource.resourceType === ResourcesTypes.Quizzes) {
     icon = resourcesData.quizzes.icon
}
```

Then we can just pass icon to child component.

Maybe I don't see better solution here. Can you please recommend a better one?

src/containers/course-sections-list/CourseSectionsList.tsx
dmtrth25: > Thanks for question. It was a tough one to fix and i struggled with it a lot. The reason I have added it is that `sectionItem` function is used inside of `map` in following code snippet:
> 
> ```js
> const sectionItems =
>     items.length === 0
>       ? clearCoorperationMenu
>       : items.map((item) => sectionsItem(item))
> ```
> 
> When I was investigating error in console, it said that the error is present inside `<SortableContext items={items} strategy={verticalListSortingStrategy}>` in the following code:
> 
> ```js
> const courseSectionContent = enabled && (
>     <>
>       <SortableContext items={items} strategy={verticalListSortingStrategy}>
>         {sectionItems}
>       </SortableContext>
>       <DragOverlay>{activeItem && sectionsItem(activeItem, true)}</DragOverlay>
>     </>
>   )
> ```
> 
> That confused me, because `SortableContext` is not being used inside any kind of loop. So I came up with solution that you can see now, because `cooperationMenu` and `SortableWrapper` are repeated on the same level and we should provide a wrapper for it to make an error disappear. I decided not no use extra `Box`, just sticked to `React`'s `Fragment`
> 
> I hope I have answered your question and it because clearer for you

I saw. Agree with you that `Box`, since a wrapper should't be here

src/containers/course-section/resource-item/ResourceItem.tsx
dmtrth25: > I found this solution rather good for this use case. I also considered following solution:
> 
> ```js
> let icon;
> 
> if (resource.resourceType === ResourcesTypes.Lessons) {
>      icon = resourcesData.lessons.icon
> }
> 
> if (resource.resourceType === ResourcesTypes.Quizzes) {
>      icon = resourcesData.quizzes.icon
> }
> ```
> 
> Then we can just pass icon to child component.
> 
> Maybe I don't see better solution here. Can you please recommend a better one?

As I see `switch case` also fits and became more readable and changeable.
For example. If we have another resource type. We can simple add new case.
What I mean

```jsx
const renderResourceIcon = () => {
  switch (resource.resourceType) {
    case ResourcesTypes.Lessons:
      return resourcesData.lessons.icon
    case ResourcesTypes.Quizzes:
      return resourcesData.quizzes.icon
    default:
      return null
  }
}
```

mhevyk: Personally I don't think so. I am pretty sure that list of resources will not change in the nearest future. Maximum that can be added are attachments, but it will be very easy to add them with current implementation.

Maybe I should just add `return null` as a default value and it will be useful improvement

dmtrth25: > just add `return null` as a default value

What do you mean add `return null` ? 
In my case, if you use switch case it's necessary to return a default value. if just to use this approach

mhevyk: I mean like following:

```jsx
const renderResourceIcon = () => {
    if (resource.resourceType === ResourcesTypes.Lessons) {
      return resourcesData.lessons.icon
    } else if (resource.resourceType === ResourcesTypes.Quizzes) {
      return resourcesData.quizzes.icon
    }
    
    return null
  }
```

dmtrth25: > I mean like following:
> 
> ```js
> const renderResourceIcon = () => {
>     if (resource.resourceType === ResourcesTypes.Lessons) {
>       return resourcesData.lessons.icon
>     } else if (resource.resourceType === ResourcesTypes.Quizzes) {
>       return resourcesData.quizzes.icon
>     }
>     
>     return null
>   }
> ```

but it's difficult to read. Are u agree ?

mhevyk: I partially agree, because for instance we can remove `else if` and refactor as following:

```jsx
const renderResourceIcon = () => {
    if (resource.resourceType === ResourcesTypes.Lessons) {
      return resourcesData.lessons.icon
    } 

   if (resource.resourceType === ResourcesTypes.Quizzes) {
      return resourcesData.quizzes.icon
    }
    
    return null
  }
```

What are your thoughts about it?

dmtrth25: > I partially agree, because for instance we can remove `else if` and refactor as following:
> 
> ```js
> const renderResourceIcon = () => {
>     if (resource.resourceType === ResourcesTypes.Lessons) {
>       return resourcesData.lessons.icon
>     } 
> 
>    if (resource.resourceType === ResourcesTypes.Quizzes) {
>       return resourcesData.quizzes.icon
>     }
>     
>     return null
>   }
> ```
> 
> What are your thoughts about it?

using switch case. It's like one line (case...) but this approach it's almost 3 lines.
This approach also make sense. 
We'll discuss it

mhevyk: I used your proposal for now after my own brainstorm after our meeting. We should rewrite cooperation activities first and then we can change it if needed in the future

dmtrth25: > I used your proposal for now after my own brainstorm after our meeting. We should rewrite cooperation activities first and then we can change it if needed in the future
> 
> Now it looks like this:
> 
> ```js
> const renderResourceIcon = () => {
>     const { Lessons, Quizzes } = ResourcesTypes
> 
>     switch (resource.resourceType) {
>       case Lessons:
>         return resourcesData.lessons.icon
>       case Quizzes:
>         return resourcesData.quizzes.icon
>       default:
>         return null
>     }
>   }
> ```

Good looking!
using destructuring for `ResourcesTypes` nice case

dmtrth25 (approved):

Mav-Ivan (approved):

byavdoshniak (approved):

PR #1927, opened 2024-05-28
Created useAccordion hook

GeorgeShvab:
- I added `useAccordion` hook and tests for it.
-  There is such property as multiple. We need it because some of our <Accordions /> can have few expanded items at the time and other can have only one expanded item at a time. There is also toggle option, which is used to control whether accordion item needs to be toggled (if the same item is clicked, should it close or still be expanded). I added it because again, different components in our project handled this logic differently.
- You may also notice that I used function overloading. Firstly I thought to use generics or conditional types, but it only made code look complicated and required few `as` keywords. Because of types, I also used Array.isArray() which can be confusing.

### DIFF
--- a/src/components/accordion-with-image/AccordionWithImage.tsx
+++ b/src/components/accordion-with-image/AccordionWithImage.tsx
@@ -1,7 +1,8 @@
-import { FC, useState } from 'react'
+import { FC } from 'react'
 
 import Accordions from '~/components/accordion/Accordions'
 import Box from '@mui/material/Box'
+import useAccordions from '~/hooks/use-accordions'
 
 import { styles } from '~/components/accordion-with-image/AccordionWithImage.styles'
 import { AccordionWithImageItem, TypographyVariantEnum } from '~/types'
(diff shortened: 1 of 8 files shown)


src/hooks/use-accordion.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+function useAccordions(
+  initialState: number[],
+  options?: Options
+): [number[], SetAccordionState]
+
+function useAccordions(
+  initialState: number,
+  options: { toggle: false }
+): [number, SetAccordionState]
```
mhevyk, line 18: Do we really need to use function overloading here? Please consider using generics instead

src/hooks/use-accordion.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+function useAccordions(
+  initialState?: number | null,
+  options?: Options
+): [number | null, SetAccordionState]
```
mhevyk, line 23: You can also consider an approach as in reducers for example, by specifying a type of accordion property. Then we can infer correct type of value, but current approach is also pretty solid for solving this problem. What do you think?

GeorgeShvab: Thanks for suggestion! I was really unsure how to properly type this hook. I made some changes and added multiple property because it is much more intuitive for other developers than previous variant. I guess it would be better to avoid function overloading and use generics with infer instead. But I still couldn't make types work properly with generics and infer 😓 (can be solved with `as` but we don't use it in our project). Maybe, we can discuss it during the meeting.

OliaHrytsak (approved):

Mav-Ivan (approved):

src/hooks/use-accordion.tsx
mhevyk: I see. Using `generics` is definitely a better option to use. As far as I know, using `any` or `as` casting is permissible inside custom hooks or utility functions, especially if they provide different signatures. The main thing is to output correct type and the typescript internals do not matter

byavdoshniak (approved):

src/hooks/use-accordion.tsx
GeorgeShvab: I replaced overloading with generics, it looks better now.

D1nviz (approved):

mhevyk (approved):

dmtrth25 (approved):

PR #1953, opened 2024-06-06
Added helper function for initials

Marichka0406:
Created helper function for receiving user's initials.

### DIFF
--- a/src/utils/helper-functions.tsx
+++ b/src/utils/helper-functions.tsx
@@ -300,3 +300,6 @@ export const parseFileName = (
     fileExtension: fileArr[fileArr.length - 1]
   }
 }
+
+export const getInitials = (firstName: string, lastName: string): string =>
+  `${firstName[0]}${lastName[0]}`.toUpperCase()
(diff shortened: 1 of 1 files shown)


D1nviz (approved):

src/utils/helper-functions.tsx
```diff
@@ -300,3 +300,6 @@ export const parseFileName = (
     fileExtension: fileArr[fileArr.length - 1]
   }
 }
+
+export const getInitials = (firstName: string, lastName: string): string =>
+  `${firstName[0]}${lastName[0]}`.toUpperCase()
```
mhevyk, line 305: Return type of this function is redundant because it can be automatically infered as string

mhevyk: What do you think about providing fallback value for `firstName` and `lastName` if they are not provided?

Marichka0406: Fixed

GeorgeShvab (approved):

mhevyk (approved):

OliaHrytsak (approved):

abalanovsky (approved):

PR #1975, opened 2024-06-20
Fixed translation of proficiency / preparation levels #1964

### LINKED ISSUE #1964: (SP: ) Fix levels of preparation translation
There is no UA translation at levels of preparation when creating the offer (choose Ukrainian language)

Знімок екрана 2024-06-17 о 17 03 05

Olenka-Hryk:
## Fixed translation of proficiency levels [ Close #1964 ]
Fixed issue with translation of `proficiency` / `preparation` levels into `Ukrainian` language.
For this, the `CheckboxList` component was extended with the ability to display custom labels passed from the outside.

| Before changing | After changing |
|-----------------|----------------|
| Before | After |

#### Improved coverage for `CheckboxList` [component](https://github.com/ita-social-projects/SpaceToStudy-Client/blob/develop/src/components/checkbox-list/CheckboxList.tsx)
_Current Test Coverage:_
Screenshot 2024-06-19 at 16 25 54

#### Added Unit Testing for `SpecializationBlock` [container](https://github.com/ita-social-projects/SpaceToStudy-Client/blob/develop/src/containers/offer-page/specialization-block/SpecializationBlock.tsx)
The following scenarios are covered in this implementation:
 - [x] **Rendering Offer Page Labels for Student:** Confirms that the correct labels are rendered for the student user role.
 - [x] **Rendering Offer Page Labels for Tutor:** Confirms that the correct labels are rendered for the tutor user role.
 - [x] **Handling Category Change:** Validates the handling of category change events.
 - [x] **Handling Category Change with `null` ID:** Ensures correct behavior when a category with null ID is selected.
 - [x] **Handling Subject Change:** Validates the handling of subject change events.
 - [x] **Handling Proficiency Level Change:** Ensures correct handling of proficiency level changes.
 - [x] **Checking `textFieldProps` for Category AsyncAutocomplete:** Validates the props of the category AsyncAutocomplete component.
 - [x] **Checking `textFieldProps` for Subject AsyncAutocomplete:** Validates the props of the subject AsyncAutocomplete component.
 
Screenshot 2024-06-20 at 00 32 47

### DIFF
--- a/src/containers/offer-page/specialization-block/SpecializationBlock.tsx
+++ b/src/containers/offer-page/specialization-block/SpecializationBlock.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useCallback } from 'react'
+import { useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import Typography from '@mui/material/Typography'
@@ -48,18 +48,6 @@ const SpecializationBlock = <T extends CreateOrUpdateOfferData>({
   const handleCheckboxesChange = (value: ProficiencyLevelEnum[]) => {
     handleNonInputValueChange('proficiencyLevel', value)
   }
-
-  const levelLabels = useMemo(
-    () =>
-      new Map(
-        Object.entries(proficiencyLevelLabels).map(([key, value]) => [
-          key,
-          t(value)
-        ])
-      ),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    []
-  )
 
   const subjectError = data.category && errors.subject
 
@@ -119,7 +107,7 @@ const SpecializationBlock = <T extends CreateOrUpdateOfferData>({
             error={t(errors.proficiencyLevel)}
             {...checkboxListProps}
             items={levelOptions}
-            labels={levelLabels}
+            labels={proficiencyLevelLabels}
             onChange={handleCheckboxesChange}
             value={data.proficiencyLevel}
           />
--- a/src/containers/offer-page/specialization-block/SpecializationBlock.tsx
+++ b/src/containers/offer-page/specialization-block/SpecializationBlock.tsx
@@ -1,4 +1,4 @@
-import { useCallback } from 'react'
+import { useMemo, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import Typography from '@mui/material/Typography'
@@ -45,7 +45,25 @@ const SpecializationBlock = <T extends CreateOrUpdateOfferData>({
   const handleCheckboxesChange = (value: ProficiencyLevelEnum[]) => {
     handleNonInputValueChange('proficiencyLevel', value)
   }
-  const levelOptions = Object.values(ProficiencyLevelEnum)
+
+  const levelOptions = useMemo(() => Object.values(ProficiencyLevelEnum), [])
+  const levelLabels = useMemo(
+    () =>
+      new Map([
+        [ProficiencyLevelEnum.Beginner, t('common.levels.beginner')],
+        [ProficiencyLevelEnum.Intermediate, t('common.levels.intermediate')],
+        [ProficiencyLevelEnum.Advanced, t('common.levels.advanced')],
+        [
+          ProficiencyLevelEnum.TestPreparation,
+          t('common.levels.testPreparation')
+        ],
+        [ProficiencyLevelEnum.Professional, t('common.levels.professional')],
+        [ProficiencyLevelEnum.Specialized, t('common.levels.specialized')]
+      ]),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
+  )
+
   const subjectError = data.category && errors.subject
 
   const checkboxListProps =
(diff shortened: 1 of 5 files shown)


YaroslavChuiko (approved):

docia (approved):

src/containers/offer-page/specialization-block/SpecializationBlock.tsx
```diff
@@ -45,7 +45,25 @@ const SpecializationBlock = <T extends CreateOrUpdateOfferData>({
   const handleCheckboxesChange = (value: ProficiencyLevelEnum[]) => {
     handleNonInputValueChange('proficiencyLevel', value)
   }
-  const levelOptions = Object.values(ProficiencyLevelEnum)
+
+  const levelOptions = useMemo(() => Object.values(ProficiencyLevelEnum), [])
```
abalanovsky, line 49: do we really need useMemo here?

Olenka-Hryk: This hook ensures that `levelOptions` won't be recreated during re-renders. In my opinion, this optimization prevents unnecessary re-renders of child components that depend on `levelOptions`.

abalanovsky: I think you can move both, ``levelOptions`` and ``levelLabels`` to constants and delete useMemo too, cause data is static

byavdoshniak: Fix it

byavdoshniak (changes requested):

src/containers/offer-page/specialization-block/SpecializationBlock.tsx
Olenka-Hryk: Done, I've fixed it.

src/containers/offer-page/specialization-block/SpecializationBlock.tsx
```diff
@@ -45,7 +48,19 @@ const SpecializationBlock = <T extends CreateOrUpdateOfferData>({
   const handleCheckboxesChange = (value: ProficiencyLevelEnum[]) => {
     handleNonInputValueChange('proficiencyLevel', value)
   }
-  const levelOptions = Object.values(ProficiencyLevelEnum)
+
+  const levelLabels = useMemo(
```
Olenka-Hryk, line 52: I moved the labels to the constants. However, this map can't be initialized outside the component context as it uses the `t` function. To not reevaluate this expression every time I applied `useMemo` hook.

Please let me know if that's okay. @abalanovsky @byavdoshniak

src/containers/offer-page/specialization-block/SpecializationBlock.tsx
```diff
@@ -45,7 +48,19 @@ const SpecializationBlock = <T extends CreateOrUpdateOfferData>({
...
+    () =>
+      new Map(
+        Object.entries(proficiencyLevelLabels).map(([key, value]) => [
+          key,
+          t(value)
+        ])
+      ),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
byavdoshniak, line 60: You are doing smth wrong if you need this in order to code work properly.

byavdoshniak (changes requested):

src/containers/offer-page/specialization-block/SpecializationBlock.tsx
abalanovsky: I think it can, you need to import **i18n** and use it like `i18n.t('common.levels.beginner')` in your constants

src/containers/offer-page/specialization-block/SpecializationBlock.tsx
Olenka-Hryk: ESLint says to add `t` into dependencies, but I initially thought that since the values in `proficiencyLevelLabels` are constants and won't change, it wasn't necessary. However, after researching, I agree with you. I still need to include `t` in the dependency array to ensure the memoized value is recomputed if necessary. Thanks for the suggestion. I'll make the change right now

src/containers/offer-page/specialization-block/SpecializationBlock.tsx
Olenka-Hryk: In that case, it will take the current language and initialize the map once. If you will change the lang - the map won't be reinitialized.

Olenka-Hryk: There is another option we can but it's less flexible. We can pass the map of codes to the `CheckboxList` component and then the checkbox will use its `t` to translate the code. Additionally, we can introduce some kind of labels options props where we can specify if the label should be treated as a text or as a code (in that case the checkbox will use the `t` to translate the code)

byavdoshniak (approved):

abalanovsky (approved):

Mike-Popovych (approved):

PR #2004, opened 2024-06-26
do not refresh the profile edit page after saving data

YaroslavChuiko:
[screen-capture (6).webm](https://github.com/ita-social-projects/SpaceToStudy-Client/assets/32570823/2fd6e4ae-d527-41d1-95ca-cf715f7ef3b8)


docia (approved):

Mygal-Alex (commented):
you need to add data update on the page 
image
if I change the data 
image
and then switch between the tabs 
and then go back to the Profile tab
I won't see the data change 
image

YaroslavChuiko:
This problem is not relevant, as it was decided to rework the logic of this component using Redux

PR #2170, opened 2024-07-25
Fix bug with disabled "Save changes" button

BohdanMylyi:
The user can click on Save changes button and save category even with empty subjects

https://github.com/user-attachments/assets/b09ef390-cc8d-4827-9f77-fad656ace63c

https://github.com/user-attachments/assets/6bc87edd-e174-4c9c-8221-3d262b02e20d

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -65,7 +65,7 @@
         "@types/react-redux": "^7.1.28",
         "@types/react-swipeable-views": "^0.13.3",
         "@types/uuid": "^10.0.0",
-        "@typescript-eslint/eslint-plugin": "^6.9.1",
+        "@typescript-eslint/eslint-plugin": "^7.17.0",
         "@typescript-eslint/parser": "^6.9.1",
         "@vitejs/plugin-react": "^4.1.1",
         "axios-mock-adapter": "^1.22.0",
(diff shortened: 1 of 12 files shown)


package-lock.json
Olenka-Hryk, line 1: Could you revert the changes in `package-lock.json`? Also, why did you update `@typescript-eslint/eslint-plugin` in this PR? Is it necessary to solve this issue? I think such updates should be in a separate PR.

Olenka-Hryk (changes requested):

package-lock.json
BohdanMylyi: > Could you revert the changes in `package-lock.json`? Also, why did you update `@typescript-eslint/eslint-plugin` in this PR? Is it necessary to solve this issue? I think such updates should be in a separate PR.

ok, I'll change it

Olenka-Hryk: Thanks 👍🏻

BohdanMylyi: > Thanks 👍🏻

i've a question, is it ok, if not all tests are passed?
because functional is right, but tests failed

Olenka-Hryk: No, it is not acceptable for **CI/tests** to fail. You can't merge a PR if there are failed tests. Several issues could be causing the failures. In your case, you changed the `AddProfessionalCategoryModal.tsx` component, which is why tests for this component are failing in CI.

I haven't investigated thoroughly, but I noticed that you modified the `const userMainSubjectTemplate` in `AddCourseTemplateModal.tsx` - `{ _id: '', name: '', appearance: { color: '', icon: '' } }` [here](src/containers/cooperation-details/add-course-modal-modal/AddCourseTemplateModal.tsx). However, the tests are still mocking it as: 
Screenshot 2024-07-25 at 17 45 01

This discrepancy might be one reason why the tests are failing. You should identify the main problem and update your tests accordingly. Tests written before the component logic change may no longer work as expected.

BohdanMylyi:
a lot of unnecessary changes

PR #2315, opened 2024-08-09
Fix coop details buttons

BohdanMylyi:
Result:
image

### DIFF
--- a/tests/unit/hooks/use-load-more.spec.jsx
+++ b/tests/unit/hooks/use-load-more.spec.jsx
@@ -38,28 +38,6 @@ describe('useLoadMore custom hook', () => {
       )
     })
   })
-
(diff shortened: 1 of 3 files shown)


Pasichnichenkoilya (approved):

dudchakk (approved):

tests/unit/hooks/use-load-more.spec.jsx
```diff
@@ -39,6 +39,27 @@ describe('useLoadMore custom hook', () => {
     })
   })
 
+  it('should open modal window', async () => {
```
abalanovsky, line 42: Why use-load-more test is here in this pull request?

docia (changes requested):
the code is good. Artem's question is also relevant for me.

PR #2342, opened 2024-08-14
Increase indent before the 'What students say' block

BohdanMylyi:
How it was:
image

Result:
image

### DIFF
--- a/src/containers/user-profile/comments-with-rating-block/CommentsWithRatingBlock.styles.ts
+++ b/src/containers/user-profile/comments-with-rating-block/CommentsWithRatingBlock.styles.ts
@@ -25,5 +25,8 @@ export const styles = {
     maxWidth: '250px',
     alignItems: 'center'
   } as SxProps<Theme>,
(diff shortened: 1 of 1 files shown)


Pasichnichenkoilya (approved):

src/containers/user-profile/comments-with-rating-block/CommentsWithRatingBlock.styles.ts
```diff
@@ -25,5 +25,8 @@ export const styles = {
     maxWidth: '250px',
     alignItems: 'center'
   } as SxProps<Theme>,
-  title: { typography: { xs: 'button', sm: 'h5', md: 'h4' } } as SxProps<Theme>
+  title: {
+    typography: { xs: 'button', sm: 'h5', md: 'h4' },
+    marginTop: '100px'
```
docia, line 30: it was decided to use a short form like `mt` and `pt`, etc for margin and padding In the project. so, it would be better to rewrite it to `mt`.

docia (approved):

dudchakk (approved):

PR #2396, opened 2024-08-27
fix location bug in user profile 

OFrankov:
Fix location bug in user profile


abalanovsky (approved):

dudchakk (changes requested):
the changes are not working, please fix that:
Screenshot 2024-08-28 121747

PR #2442, opened 2024-09-04
Photo step - TS bugfix

OliaHrytsak:
https://github.com/user-attachments/assets/3d14f2b4-9438-467d-a3d3-e25393a85ef2

### DIFF
--- a/src/types/common/interfaces/common.interfaces.ts
+++ b/src/types/common/interfaces/common.interfaces.ts
@@ -172,12 +172,18 @@ export interface CourseSectionHandlers {
   titleText?: string
 }
 
+declare global {
+  export interface File {
+    src?: string
+  }
+}
+
 export interface StepData {
   generalInfo: {
     data: UserGeneralInfo
     errors: Record<string, string>
   }
-  photo: string[]
+  photo?: File[] | undefined
   subjects: Array<SubjectNameInterface & { category: CategoryNameInterface }>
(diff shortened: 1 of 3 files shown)


BohdanMylyi:
Fix sonar issues
image

docia (approved):

src/types/common/interfaces/common.interfaces.ts
```diff
@@ -172,12 +172,18 @@ export interface CourseSectionHandlers {
   titleText?: string
 }
 
+declare global {
+  export interface File {
+    src?: string
+  }
+}
```
YaroslavChuiko, line 179: Why do we need this optional `src` property on the global File type, if the File object by default has a `name` property that can be a string representing the file name or the path to the file? [MDN File() constructor documentation](https://developer.mozilla.org/en-US/docs/Web/API/File/File#filename)

BohdanMylyi (approved):

src/types/common/interfaces/common.interfaces.ts
OliaHrytsak: The optional src property on the global File type is introduced in this code to hold the image's display URL after it has been resized or processed

src property is used for convenience in dealing with image previews or resized images in the application, allowing the file metadata to remain intact while also providing an accessible path for rendering the image

YaroslavChuiko (approved):

PR #2596, opened 2024-10-10
Implement user online-offline status

### LINKED ISSUE #2570: (SP: 5) Implement user online-offline status
## Description
Implement real user online-offline statuses on the chat page using sockets instead of mock data.

YaroslavLys:
## Changes

- Implemented socket connection with redux middleware (best practice according to the [official docs](https://redux.js.org/faq/code-structure#where-should-websockets-and-other-persistent-connections-live))
- Implemented socketSlice
- Implemented creation of the socket instance using singleton pattern
- Implemented socket events to track users' status
- Added badges to show user's status
- Covered new lines of code with tests
- Fixed old tests

## Demo

https://github.com/user-attachments/assets/aa29c380-13e1-4671-8639-869ef7fedc6e


src/components/user-profile-info/UserProfileInfo.tsx
```diff
@@ -50,10 +50,10 @@ const UserProfileInfo: FC<UserProfileInfoProps> = ({
   sx = {},
   _id,
   role,
-  renderAdditionalInfo = true,
-  onlineBadge = false
+  renderAdditionalInfo = true
 }) => {
   const { t } = useTranslation()
+  const { usersOnline } = useAppSelector((state) => state.socket)
```
myevd: Consider doing this in the parent component, so every `usersOnline` change doesn't trigger re-render

src/redux/features/socketSlice.ts
```diff
@@ -0,0 +1,34 @@
...
+  initialState,
+  reducers: {
+    connectSocket: (state) => {
+      state.isConnected = true
+    },
+    disconnectSocket: () => {
+      return initialState
+    },
+    setUsersOnline: (state, action: PayloadAction<string[]>) => {
+      state.usersOnline = action.payload
```
myevd, line 25: Is there a chance that action.payload will be the same as the actual usersOnline list? If so I would consider a comparison before dispatch

myevd: If SocketEvent.UsersOnline will be triggered very frequently, a debouncing or throttling could be applied(maybe right now it's not a case, but for future)

src/components/user-profile-info/UserProfileInfo.tsx
YaroslavLys: To optimise re-renders instead of moving `useAppSelector` to the parent component I decided to create memoized selector that only checks if the specific user is online. This will only trigger re-renders for the component when the online status of the specific user (id) changes, instead of whenever the `usersOnline` array is modified. This selector I put in the `redux/selectors` folder. As a result I managed to optimise re-renders of `UserProfileInfo`, `ChatItem` and `ChatHeader` components in my latest commit.

src/redux/features/socketSlice.ts
YaroslavLys: That's a good idea to apply debouncing to the `setUsersOnline`! In my latest commit I wrote my custom debounce function similar to the one in `loadsh` library. I used debounce in a `socket-middleware`. It reduced the frequency of `setUsersOnline` updates and improved performance by limiting how often the store is updated (once per second).

src/containers/chat/chat-item/ChatItem.tsx
```diff
@@ -92,6 +94,8 @@ const ChatItem: FC<ItemOfChatProps> = ({
           badgeContent={
             <Typography component={ComponentEnum.Span} sx={styles.active} />
           }
+          // invisible={!isOnline}
```
Mav-Ivan: remove this comment)

myevd (approved):

YaroslavChuiko (approved):

PavloDolia (approved):

dudchakk (approved):

BohdanMylyi (approved):

PR #2598, opened 2024-10-10
Added refreshing for an access token

### LINKED ISSUE #2574: Fix refresh logic for access tokens
When an access token expires, current refresh logic can't handle this situation and user gets a response with 401 code
Screenshot 2024-10-04 at 13 17 16
(checkAuth is an action creator, that calls the refresh API endpoint)
Screenshot 2024-10-04 at 13 26 50

PavloDolia:
- [x] addedd refresh logic for an access token

If you want to test it by reducing the lifetime of an access token, please don't set it to less than 2 seconds because it will cause conflicts in refresh logic if multiple requests are sent on a page at the same time

### DIFF
--- a/src/services/setup-interceptors.ts
+++ b/src/services/setup-interceptors.ts
@@ -1,12 +1,11 @@
-import {
-  AxiosError,
-  InternalAxiosRequestConfig,
-  AxiosHeaders,
-  AxiosResponse
-} from 'axios'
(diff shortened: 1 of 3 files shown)


dudchakk (approved):
great job!

Mav-Ivan (approved):

ipasic-softserve (approved):

PR #2627, opened 2024-10-16
Fixed updating offer count when creating new offer in CompleteProfileBlock

### LINKED ISSUE #2606: (SP: 2) Fix updating offer data in CompleteProfileBlock
Fix this bug:
- [x] Offer ProfileItem is only updated when user creates new offer **and refreshes the page**: is noticable in PR #2605

dudchakk:
**How it works now:**

https://github.com/user-attachments/assets/153e0a9d-53b1-4b1d-8f64-ff4b0955a508

### DIFF
--- a/src/components/complete-profile/CompleteProfileBlock.tsx
+++ b/src/components/complete-profile/CompleteProfileBlock.tsx
@@ -24,6 +24,9 @@ import { styles } from '~/components/complete-profile/CompleteProfileBlock.style
 import useAxios from '~/hooks/use-axios'
 import { OfferService } from '~/services/offer-service'
 import { defaultResponse } from '~/pages/my-offers/MyOffers.constants'
+import { useDrawer } from '~/hooks/use-drawer'
+import AppDrawer from '~/components/app-drawer/AppDrawer'
+import CreateOffer from '~/containers/offer-page/create-offer/CreateOffer'
 
 interface CompleteProfileBlockProps {
   profileItems: ProfileItemType[]
@@ -42,6 +45,10 @@ const CompleteProfileBlock: FC<CompleteProfileBlockProps> = ({
   const homePage = useMatch(guestRoutes[userRole as UserRole].path)
   const [isOpen, setIsOpen] = useState(false)
 
+  const { openDrawer, closeDrawer, isOpen: isDrawerOpen } = useDrawer()
+  const [isOfferCreated, setIsOfferCreated] = useState(false)
+  const handleOpenDrawer = () => openDrawer()
+
   useEffect(() => {
     if (openAccordion) {
       setIsOpen(true)
@@ -53,7 +60,7 @@ const CompleteProfileBlock: FC<CompleteProfileBlockProps> = ({
       OfferService.getUsersOffers({
         id: userId
       }),
-    [userId]
+    [userId, isOfferCreated]
   )
 
   const { response } = useAxios({
@@ -99,6 +106,7 @@ const CompleteProfileBlock: FC<CompleteProfileBlockProps> = ({
     () =>
       profileItems.map((item) => (
         <ProfileItem
+          handleOpenDrawer={handleOpenDrawer}
           isFilled={checkProfileData.includes(item)}
           item={item}
           key={item.id}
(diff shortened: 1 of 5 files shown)


BohdanMylyi (approved):
👍

src/components/complete-profile/CompleteProfileBlock.tsx
```diff
@@ -53,7 +60,7 @@ const CompleteProfileBlock: FC<CompleteProfileBlockProps> = ({
       OfferService.getUsersOffers({
```
PavloDolia: you don't have to use the useCallback hook at all, you can pass the `OfferService.getUsersOffers` function to the useAxios hook, or at least you should delete `isOfferCreated` from the dependancy array.

src/components/complete-profile/CompleteProfileBlock.tsx
```diff
@@ -42,6 +45,10 @@ const CompleteProfileBlock: FC<CompleteProfileBlockProps> = ({
   const homePage = useMatch(guestRoutes[userRole as UserRole].path)
   const [isOpen, setIsOpen] = useState(false)
 
+  const { openDrawer, closeDrawer, isOpen: isDrawerOpen } = useDrawer()
+  const [isOfferCreated, setIsOfferCreated] = useState(false)
+  const handleOpenDrawer = () => openDrawer()
```
PavloDolia: why did you create `handleOpenDrawer` function? You don' have to do it, and the `handleOpenDrawer` function will be recreated after each render, which can affect productivity. As far as I understand you want to rename the `openDrawer` function, so you could destructure with a renaming, like in [this example](https://stackoverflow.com/questions/57065348/destructuring-and-rename-property)

src/components/profile-item/ProfileItem.tsx
```diff
@@ -8,29 +8,25 @@ import { styles } from '~/components/profile-item/ProfileItem.styles'
...
   userRole: UserRoleEnum | ''
+  handleOpenDrawer?: () => void | undefined
 }
 
 const ProfileItem = ({
   item,
   userRole,
-  isFilled = false
+  isFilled = false,
+  handleOpenDrawer = undefined
```
PavloDolia: we don't have to set `undefined` as a default value for an object property, because it will be `undefined` if the object doesn't contain a property

src/containers/offer-page/create-offer/CreateOffer.tsx
```diff
@@ -8,16 +8,26 @@ import { CreateOrUpdateOfferData } from '~/types'
 
 interface CreateOfferProps {
   closeDrawer: () => void
+  updateOffer?: Dispatch<SetStateAction<boolean>>
 }
 
-const CreateOffer: FC<CreateOfferProps> = ({ closeDrawer }) => {
+const CreateOffer: FC<CreateOfferProps> = ({
+  closeDrawer,
+  updateOffer = undefined
```
PavloDolia: don't need to set a property to `undefined` here

src/containers/offer-page/create-or-edit-offer/CreateOrEditOffer.tsx
```diff
@@ -38,15 +38,17 @@ import { openAlert } from '~/redux/features/snackbarSlice'
...
   service: ServiceFunction<Offer | null, CreateOrUpdateOfferData>
+  updateOffer?: Dispatch<SetStateAction<boolean>>
 }
 
 const CreateOrEditOffer: FC<CreateOrUpdateOfferProps> = ({
   existingOffer = null,
   closeDrawer,
-  service
+  service,
+  updateOffer = undefined
```
PavloDolia: `undefined` isn't needed

PavloDolia (changes requested):

src/components/complete-profile/CompleteProfileBlock.tsx
dudchakk: I created this function following the example at the MyOffers.tsx component. But looks like you are right, I’ll try to remove it

src/components/complete-profile/CompleteProfileBlock.tsx
dudchakk: I use isOfferCreated dependency because it makes the hook re-send request to get user offers when the new offer is created in CompleteProfileBlock. It was the main task) without this dependency the first created offer won’t be shown in CompleteProfileBlock unless the user refreshes the page

PavloDolia: ok, now I understand your idea, but we should use the `useEffect` hook for such purposes because it will make code much more readable

PavloDolia (changes requested):

src/components/complete-profile/CompleteProfileBlock.tsx
PavloDolia: ok, now I understand your idea, but we should use the `useEffect` hook if you want to re-send request because it will make code much more readable

PavloDolia (approved):

ipasic-softserve (approved):

Mav-Ivan (approved):

PR #2628, opened 2024-10-17
Enhanced chat search

### LINKED ISSUE #2610: (SP: 2) Enhance chat search
The current chat search functionality only works with the messages that are currently loaded, not across all messages:

image
image

This needs to be updated so that the search covers all messages, both loaded and unloaded.

Example: If a user searches for the term "test" and there is one match in the 10 currently loaded messages, but another match exists 40 messages back, the system should load all 40 messages and display the relevant matches.

Mav-Ivan:
https://github.com/user-attachments/assets/2056631e-1fe4-4431-a199-b61c3b72247d

### DIFF
--- a/src/containers/chat/chat-header/ChatHeader.tsx
+++ b/src/containers/chat/chat-header/ChatHeader.tsx
@@ -1,4 +1,13 @@
-import { FC, MouseEvent, useRef, useState } from 'react'
+import {
+  FC,
+  MouseEvent,
+  SetStateAction,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+  Dispatch
+} from 'react'
 import { useTranslation } from 'react-i18next'
 import Box from '@mui/material/Box'
 import Typography from '@mui/material/Typography'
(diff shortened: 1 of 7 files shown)


ipasic-softserve (approved):

src/containers/chat/chat-header/ChatHeader.tsx
```diff
@@ -76,6 +96,32 @@ const ChatHeader: FC<ChatHeaderProps> = ({
...
+  const getAllMessages = useCallback(
+    () =>
+      messageService.getMessages({
+        chatId: currentChat._id,
+        limit: 100,
+        skip: 0
+      }),
+    [currentChat._id]
+  )
+
```
dudchakk, line 108: why did you omit `setAllMessages` in the dependancy array?

Mav-Ivan: thanks for noticing)

dudchakk (approved):

src/containers/chat/chat-header/ChatHeader.tsx
```diff
@@ -76,6 +96,35 @@ const ChatHeader: FC<ChatHeaderProps> = ({
     setIsSearchOpen(false)
   }
 
+  const getAllMessages = useCallback(
+    () =>
+      messageService.getMessages({
+        chatId: currentChat._id,
+        limit: 100,
```
PavloDolia: why do you fetch only 100 messages? `Number.MAX_SAFE_INTEGER` will be better I guess

PavloDolia (commented):
as far as I understand, searching for messages should happen on the backend, but you should ask our tech expert about it, maybe he will say this approach is alright for our app

src/containers/chat/chat-header/ChatHeader.tsx
Mav-Ivan: fixed

Mike-Popovych (approved):

PR #2630, opened 2024-10-18
Display active tab in URL on "My cooperations" and "My offers" pages

### LINKED ISSUE #2629: (SP: 2) Display active tab in URL on "My cooperations" and "My offers" pages
- [ ] Active tab on the "My cooperations" and "My offers" pages should be displayed in the URL

YaroslavChuiko:
## Description
This PR adds functionality to display the active tab in the URL on the **"My Cooperations"** and **"My Offers"** pages. Now, when users navigate between tabs, the URL updates reflect the selected tab, allowing for improved navigation and sharing specific tab views.

## Changes Made
- Updated routing logic to capture and display the active tab as a URL query parameter.
- Moved tabs into new component TabFilterList to reuse it across different pages
- Modified tab component behavior to sync with the URL, ensuring the correct tab is shown when visiting a URL with an active tab parameter.
- Handling for default tab selection was added when no tab parameter was present.

## Testing
- Verified that switching tabs updates the URL correctly.
- Confirmed that reloading the page with an active tab in the URL correctly displays the corresponding tab.

### DIFF
--- a/src/pages/my-cooperations/MyCooperations.tsx
+++ b/src/pages/my-cooperations/MyCooperations.tsx
@@ -1,6 +1,6 @@
-import { useCallback, useLayoutEffect, useState } from 'react'
+import { useCallback, useEffect, useLayoutEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
-import { Link } from 'react-router-dom'
+import { Link, useSearchParams } from 'react-router-dom'
 import Box from '@mui/material/Box'
 import Typography from '@mui/material/Typography'
 
@@ -11,7 +11,6 @@ import useAxios from '~/hooks/use-axios'
 import useSort from '~/hooks/table/use-sort'
 import useBreakpoints from '~/hooks/use-breakpoints'
 import useFilter from '~/hooks/table/use-filter'
-import Tab from '~/components/tab/Tab'
 import Loader from '~/components/loader/Loader'
 import AppButton from '~/components/app-button/AppButton'
 import AppPagination from '~/components/app-pagination/AppPagination'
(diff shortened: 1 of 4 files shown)


PavloDolia (approved):

ipasic-softserve (approved):

src/pages/my-cooperations/MyCooperations.tsx
dudchakk, line 1: you can move the type definition from this file to one in the 'types' container

src/pages/my-offers/MyOffers.tsx
dudchakk, line 1: same with type

src/pages/my-offers/MyOffers.tsx
```diff
@@ -56,6 +64,10 @@ const MyOffers = () => {
 
   const handleOpenDrawer = () => openDrawer()
```
dudchakk, line 66: add brackets for `if` block for readabitily, unless we have a convention to omit them if the `if` block has only one statement

dudchakk (changes requested):

src/pages/my-cooperations/MyCooperations.tsx
YaroslavChuiko: I think we shouldn't move this type definition to a separate file because this type is specific for MyCooperation and is used only in this component

src/pages/my-offers/MyOffers.tsx
YaroslavChuiko: I think we shouldn't move this type definition to a separate file because this `type` is specific for MyOffers and is used only in this component

src/pages/my-cooperations/MyCooperations.tsx
dudchakk: Okay, I see that this type depends on `tabsInfo`, so we can leave its definition here

dudchakk (approved):

PR #2692, opened 2024-11-13
fixed the email input erroring

### LINKED ISSUE #2671: (SP: 1) Fix google auth
When users sign in with Google account, they see the email error `This field cannot be empty`, but they shouldn't see this error.

markgol777:
Removed autoFocus from one of the input fields in order to avoid triggering validation on button click, as the button click would lead to a blur event, thereby causing the validation to activate.

How it looks now:
image


ipasic-softserve (approved):
Don't forget to add yourself to the assignees and link the issue in the development section.
image

Mav-Ivan (approved):

Renatavl (approved):

PR #2834, opened 2024-11-21
Added reusable Switch Component according to design systems

amoutens:
Added switch aligned with design systems [](https://www.figma.com/design/liJZDYhUnDP15Pi9bipDcv/Space2Study-(Students)?node-id=20967-98387&node-type=frame&t=2vrL8A5QMjqbrSaJ-0)

Added optional props that include:
`label`
`labelPosition`
`size`
`loading`
`disabled`

default:
image

checked:
image

hover:
image

image

loading and disabled:
image

label and label positions:
image
image
image
image

sizes change (medium by default):
Large:
image
Small:
image

Have some problems with disabled hover, please contact my tg @amoutens

### DIFF
--- a/src/components/app-switch/AppSwitch.styles.ts
+++ b/src/components/app-switch/AppSwitch.styles.ts
@@ -7,10 +7,7 @@ const trackMixin = (width: number, borderWidth: number) => ({
   borderColor: 'primary.200',
   borderRadius: `${width / 3}px`,
   opacity: '1 !important',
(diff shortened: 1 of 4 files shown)


nebby2105:
It might be better to move the AppSwitch component to the design-system folder, along with its styles and stories

Mav-Ivan (changes requested):
move component to src/design-system/components
move stories to src/design-system/stories
move test to tests/unit/design-system
Also use aliases `~scss-components `, `~scss`

PR #2847, opened 2024-11-22
Fix tutor homepage schedule localization

### LINKED ISSUE #2845: [Tutor Home] The schedule block isn't translated into Ukrainian.
**Description:**
The schedule block on the Tutor homepage is not displaying text in Ukrainian, even though the user's interface language is set to Ukrainian

**Environment:**
OS - Windows 11
Browser - Chrome 131.0.6778.86
Reproducible: always.

**Precondition:**
1.User is logged In as a Tutor on https://s2s-front-stage.azurewebsites.net/

**Steps to reproduce:**
Open the “Home” page.
Scroll to "Your schedule" block.

**Actual Result:**
The schedule block text is not translated into Ukrainian at the Tutor Home page.
image

**Expected Result:**
The schedule block text is translated into Ukrainian at the Tutor Home page.

sandrvvu:
**Problem:** The schedule block on the Tutor homepage is not displaying text in Ukrainian, even though the user's interface language is set to Ukrainian.
image

**Result:**
(UA)
image
(EN)
image


src/components/tutor-schedule/TutorSchedule.tsx
```diff
@@ -33,19 +34,25 @@ const items: ITutorScheduleItem[] = [
...
+        </Typography>
       </Box>
       <Box sx={styles.cardWrapper}>
         {items.map((item) => (
           <TutorScheduleCard item={item} key={item.lastName} />
         ))}
       </Box>
       <Button sx={styles.btn} variant='text'>
-        All scheduled classes
+        {t('tutorHomePage.schedule.AllClasses')}
```
Made1ra: I think it's better to use lower camel case here.
E.g. `allClasses` instead of `AllClasses`.

Made1ra (changes requested):

src/components/tutor-schedule/TutorSchedule.tsx
sandrvvu: oh right, my bad)

Made1ra (approved):

dudchakk (changes requested):
probably the price and date should be translated as well)
in this example it's `80 UAH/hour` and `18:00-19:30 Thu`

Renatavl (approved):

Made1ra (approved):

markgol777 (approved):

dudchakk (approved):

PR #2850, opened 2024-11-22
added save the password disabling

markgol777:
image
If there's nothing entered the button will be disabled


ArtemHolikov (approved):

luiqor (approved):

src/containers/edit-profile/password-security-tab/change-password-modal/ChangePasswordModal.tsx
```diff
@@ -196,6 +194,9 @@ const ChangePasswordModal = () => {
               {t('common.cancel')}
             </AppButton>
             <AppButton
+              disabled={
+                !data.password || !data.confirmPassword || !data.currentPassword
```
ShadowOfTheSpace, line 198: Extract this statement to variable. It will increase readability here

PR #2908, opened 2024-11-29
Added Icon Button Component

### LINKED ISSUE #2701: (SP: 5) Create IconButton component aligned with design system
#### Description:
Implement a reusable `IconButton` component that meets the design system standards. This component will be used throughout the application to maintain consistency in icon button styles, behavior, and accessibility.

[Figma](https://www.figma.com/design/liJZDYhUnDP15Pi9bipDcv/Space2Study-(Students)?node-id=20967-96971&t=l5PI8JRXpNFKQBr0-4)

#### Requirements:
- **Styles:**
  - Follow design system variables for colors, typography, padding, and other styles.
  - Support different variants (eg., primary, secondary).
  - Ensure hover, focus, active, and disabled states are defined according to the design system.
  
- **Props:**
  - `variant`: Defines type (primary, secondary, etc.).
  - `size`: Supports different sizes (xs, sm, md, lg).
  - `loading`: Disables the icon button when true. When loading, the  icon button shouldn't shrink or change its size compared to its default state
  - `disabled`: Disables the icon button when true.

Don't forget to cover component with tests and add to the storybook

amoutens:
Added IconButton according to design system with the props below:
`variant`
`size`
`loading`
`disabled`
`toggleAble`

All Variant Default:
image
Hover:
image
Focused:
image
Disabled:
image
Loading:
image
ToggleAble:
image


src/design-system/components/icon-button/IconButton.tsx
```diff
@@ -0,0 +1,63 @@
+import { CircularProgress, IconButtonProps } from '@mui/material'
+import React, { useState } from 'react'
+import { IconButtonVariant } from '~/types'
+import AddRoundedIcon from '@mui/icons-material/AddRounded'
+import { IconButton as MuiIconButton } from '@mui/material'
+import './IconButton.scss'
+interface S2SIconButtonProps extends Omit<IconButtonProps, 'size'> {
```
ShadowOfTheSpace, line 12: Add empty line between imports and interface declaration

```suggestion

interface S2SIconButtonProps extends Omit<IconButtonProps, 'size'> {
```

src/design-system/components/icon-button/IconButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const classNamesContainerIconBG = [
+    's2s-icon-button',
+    `s2s-icon-button--${size}`,
+    `s2s-icon-button--${variant}${isToggled ? '-toggle-able' : ''}`
+  ].join(' ')
+  const classNamesContainerIcon = [
+    's2s-icon',
+    `s2s-icon--${size}`,
+    `s2s-icon--${variant}${isToggled ? '-toggle-able' : ''}`
+  ].join(' ')
```
ShadowOfTheSpace: Use `cn` function here

src/design-system/components/icon-button/IconButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+interface S2SIconButtonProps extends Omit<IconButtonProps, 'size'> {
+  variant?: IconButtonVariant
+  size?: 'xs' | 'sm' | 'md' | 'lg'
+  loading?: boolean
+  disabled?: boolean
+  toggleAble?: boolean
+}
+
+export const IconButton: React.FC<S2SIconButtonProps> = ({
+  variant = IconButtonVariant.Primary,
```
ShadowOfTheSpace, line 23: I am not sure, but I think you should not use enums which are outside of design-system folder, because this basic component will be moved to another repository later.

ShadowOfTheSpace (changes requested):

src/design-system/components/icon-button/IconButton.tsx
amoutens: To be more consistent with existing solutions I think it is better to use common.enums.ts. For better approach I can move  my enum to design-system folder.

src/design-system/components/icon-button/IconButton.tsx
amoutens: Thanks, didn`t notice it!

ShadowOfTheSpace (approved):

SofiiaYevush (approved):

src/design-system/components/icon-button/IconButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+  size = 'md',
+  loading = false,
+  disabled = false,
+  toggleAble = false,
+  ...props
+}) => {
+  const [isToggled, setIsToggled] = useState<boolean>(false)
```
PavloButynets: All basic components must be stateless and all logic must be implemented at the level of the parent component (the value and the function to change this value must be passed through props)

PavloButynets (changes requested):

src/design-system/components/icon-button/IconButton.tsx
amoutens: According to my issue condition, there should not be state values ​​in props and the code will not be as clean because of this solution

markgol777 (approved):

src/design-system/components/icon-button/IconButton.tsx
PavloButynets: Then you can edit the description of the issue, here the component is quite limited because handleClick only changes the behavior of the component itself, but this function can perform more difficult logic (for example, API request, interaction with other components or states, etc.). That is why the component should not have its own state and any other logic,
because now the only thing it can do is simply change the state of isTogled and make the appropriate style changes

PavloButynets (changes requested):

nebby2105 (approved):

PavloButynets (approved):

PR #2912, opened 2024-11-29
Replaced enums from all types to design system folder

### LINKED ISSUE #2911: (SP: 0.5) Replace all enums related to design system components
Need to replace all enums from common.enums.ts related to design system components into separate constants files

amoutens:
Created constants.ts files for all components in design-system, who have enums in common.enum.ts in types folder and replace enums to this constants

### DIFF
--- a/src/design-system/components/input-field/InputField.tsx
+++ b/src/design-system/components/input-field/InputField.tsx
@@ -1,5 +1,5 @@
 import '~scss-components/input-field/InputField.scss'
-import { InputFieldVariantEnum } from '~/types'
+import { InputFieldVariantEnum } from './InputField.constants'
 import SearchIcon from '@mui/icons-material/Search'
 import ClearIcon from '@mui/icons-material/Clear'
 import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline'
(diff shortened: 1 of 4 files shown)


src/design-system/components/input-field/InputField.tsx
```diff
@@ -1,5 +1,5 @@
 import '~scss-components/input-field/InputField.scss'
-import { InputFieldVariantEnum } from '~/types'
+import { InputFieldVariantEnum } from './InputField.constants'
```
ShadowOfTheSpace: Move this import to the bottom of imports. It's an excellent practice when all imports are not chaotically sorted.
Firstly - all imports from external libs, after that - imports from global packages in our project, and last ones - from neighboring files.
Unfortunately, we don't have a linter for that, but I think we should add it in the future.

ShadowOfTheSpace (approved):

Made1ra (approved):

src/design-system/components/input-field/InputField.tsx
amoutens: Fixed!)

ArtemHolikov (approved):

PR #2974, opened 2024-12-12
Made checkbox controlled

### LINKED ISSUE #2970: Make Checkbox component controllable
Currently, the `CheckBox` component is designed as an uncontrolled component, meaning its checked state is managed internally within the component using useState. This makes it impossible to control the checkbox state from outside the component.
**Expected Behavior:** We need to refactor the `CheckBox` component to support both controlled and uncontrolled states. The state should be able to be managed externally through props (e.g., `checked`, `onChange`) and internally as a fallback when the checked prop is not provided.

`Proposed Solution:`

- Add a `checked` prop to the component to allow external control of the checkbox's checked state.
- Add an `onChange` callback prop that will be triggered when the checkbox state changes.
- If the `checked` prop is provided, use it as the source of truth for the checkbox state; otherwise, fall back to using the internal state (with useState).
- Ensure the component still works correctly when used in an uncontrolled manner by managing the state internally when `checked` is not provided.
- Update storybook to reflect these changes

nebby2105:
Refactored the CheckBox component to support both controlled and uncontrolled states. #2970 
The changes allow external state management via props (`checked` and `onChange`) while preserving internal state handling as a fallback when the `checked` prop is not provided.

Here’s a video demonstrating how the controlled CheckBox interacts with external state:

https://github.com/user-attachments/assets/359751f5-0512-4eb9-b9e9-f31b599ef41e

Summary:

- [x] Added a `checked` prop to enable external control of the checkbox state.
- [x] Introduced an `onChange` callback prop to handle state change events.
- [x] Implemented fallback logic to use internal state when the `checked` prop is not passed.
- [x] Preserved support for uncontrolled usage with the `defaultChecked` prop.
- [x] Added a new "Controlled" story to demonstrate the controlled behavior of the CheckBox component.
- [x] Updated component descriptions to reflect support for both controlled and uncontrolled states.

### DIFF
--- a/src/design-system/components/checkbox/CheckBox.tsx
+++ b/src/design-system/components/checkbox/CheckBox.tsx
@@ -1,5 +1,5 @@
 import Checkbox, { CheckboxProps } from '@mui/material/Checkbox'
-import { type ChangeEvent, ReactNode, useCallback, useState } from 'react'
+import { type ChangeEvent, type ReactNode, useCallback, useState } from 'react'
 import Loader from '~/components/loader/Loader'
 import { cn } from '~/utils/cn'
 
@@ -12,12 +12,14 @@ interface CheckBoxProps extends Omit<CheckboxProps, 'size' | 'onChange'> {
   color?: 'primary' | 'secondary' | 'error' | 'success'
   size?: 'sm' | 'md' | 'lg'
   loading?: boolean
-  isChecked?: boolean
+  checked?: boolean
+  defaultChecked?: boolean
   onChange?: (checked: boolean) => void
 }
 
 const CheckBox: React.FC<CheckBoxProps> = ({
-  isChecked: externalIsChecked,
+  checked: externalIsChecked,
+  defaultChecked = false,
   onChange,
   color = 'primary',
   disabled = false,
@@ -28,7 +30,9 @@ const CheckBox: React.FC<CheckBoxProps> = ({
   size = 'md',
   ...props
 }) => {
-  const [internalIsChecked, setInternalIsChecked] = useState<boolean>(false)
+  const [internalIsChecked, setInternalIsChecked] = useState<boolean>(
+    defaultChecked ?? false
+  )
 
   const isCheckboxControllable = externalIsChecked !== undefined
 
--- a/src/design-system/components/checkbox/CheckBox.tsx
+++ b/src/design-system/components/checkbox/CheckBox.tsx
@@ -12,9 +12,16 @@ interface CheckBoxProps extends Omit<CheckboxProps, 'size'> {
   color?: 'primary' | 'secondary' | 'error' | 'success'
   size?: 'sm' | 'md' | 'lg'
   loading?: boolean
+  checked?: boolean
+  onChange?: (
+    event: React.ChangeEvent<HTMLInputElement>,
+    checked: boolean
+  ) => void
 }
 
 const CheckBox: FC<CheckBoxProps> = ({
+  checked,
+  onChange,
   color = 'primary',
   disabled = false,
   label,
@@ -24,10 +31,18 @@ const CheckBox: FC<CheckBoxProps> = ({
   size = 'md',
   ...props
 }) => {
-  const [checked, setChecked] = useState<boolean>(false)
+  const [internalChecked, setInternalChecked] = useState<boolean>(false)
+
+  const isCheckboxControllable = checked !== undefined
 
   const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setChecked(event.target.checked)
+    const newChecked = event.target.checked
+
+    if (!isCheckboxControllable) {
+      setInternalChecked(newChecked)
+    }
+
+    onChange?.(event, newChecked)
   }
 
   const loaderSizeMapping: Record<string, number> = {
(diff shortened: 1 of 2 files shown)


src/design-system/components/checkbox/CheckBox.tsx
```diff
@@ -12,9 +12,16 @@ interface CheckBoxProps extends Omit<CheckboxProps, 'size'> {
   color?: 'primary' | 'secondary' | 'error' | 'success'
   size?: 'sm' | 'md' | 'lg'
   loading?: boolean
+  checked?: boolean
+  onChange?: (
+    event: React.ChangeEvent<HTMLInputElement>,
```
ShadowOfTheSpace: We don't need to pass this event to the `onChange` callback

src/design-system/components/checkbox/CheckBox.tsx
```diff
@@ -24,10 +31,18 @@ const CheckBox: FC<CheckBoxProps> = ({
   size = 'md',
   ...props
 }) => {
-  const [checked, setChecked] = useState<boolean>(false)
+  const [internalChecked, setInternalChecked] = useState<boolean>(false)
```
ShadowOfTheSpace: Add `defaultIsChecked` prop and use it here. And also rename state to `internalIsChecked`, please
```suggestion
  const [internalChecked, setInternalChecked] = useState<boolean>(defaultIsChecked ?? false)
```

src/design-system/components/checkbox/CheckBox.tsx
```diff
@@ -12,9 +12,16 @@ interface CheckBoxProps extends Omit<CheckboxProps, 'size'> {
...
   loading?: boolean
+  checked?: boolean
+  onChange?: (
+    event: React.ChangeEvent<HTMLInputElement>,
+    checked: boolean
+  ) => void
 }
 
 const CheckBox: FC<CheckBoxProps> = ({
+  checked,
```
ShadowOfTheSpace: `checked` is a boolean, so rename it to `isChecked`. But also, to be more consistent in naming in this component, you should use rename while destructuring.
So you will have:
```isChecked: externalIsChecked```

File: src/design-system/components/checkbox/CheckBox.tsx [hunk 4]
```diff
@@ -24,10 +31,18 @@ const CheckBox: FC<CheckBoxProps> = ({
...
   ...props
 }) => {
-  const [checked, setChecked] = useState<boolean>(false)
+  const [internalChecked, setInternalChecked] = useState<boolean>(false)
+
+  const isCheckboxControllable = checked !== undefined
 
   const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    setChecked(event.target.checked)
+    const newChecked = event.target.checked
```
ShadowOfTheSpace: We don't need event in `onChange` callback, so you can use destructurization here or even in function declaration, like `{ target: { checked } } : React.ChangeEvent<HTMLInputElement>)`

src/design-system/components/checkbox/CheckBox.tsx
```diff
@@ -37,6 +52,7 @@ const CheckBox: FC<CheckBoxProps> = ({
   }
 
   const loader = <Loader size={loaderSizeMapping[size]} />
+  const resolvedCheck = isCheckboxControllable ? checked : internalChecked
```
ShadowOfTheSpace:
```suggestion
  const resolvedIsChecked = isCheckboxControllable ? externalIsChecked: internalIsChecked
```

src/design-system/components/checkbox/CheckBox.tsx
```diff
@@ -60,12 +76,12 @@ const CheckBox: FC<CheckBoxProps> = ({
       ) : (
         <Checkbox
           {...props}
-          checked={checked}
+          checked={resolvedCheck}
```
ShadowOfTheSpace:
```suggestion
          checked={resolvedIsChecked}
```

ShadowOfTheSpace (changes requested):

src/design-system/components/checkbox/CheckBox.tsx
```diff
@@ -1,20 +1,24 @@
 import Checkbox, { CheckboxProps } from '@mui/material/Checkbox'
-import { FC, ReactNode, useState } from 'react'
+import { type ChangeEvent, ReactNode, useCallback, useState } from 'react'
```
ShadowOfTheSpace:
```suggestion
import { type ChangeEvent, type ReactNode, useCallback, useState } from 'react'
```

ShadowOfTheSpace (changes requested):

src/design-system/components/checkbox/CheckBox.tsx
```diff
@@ -24,11 +30,22 @@ const CheckBox: FC<CheckBoxProps> = ({
   size = 'md',
   ...props
 }) => {
-  const [checked, setChecked] = useState<boolean>(false)
+  const [internalIsChecked, setInternalIsChecked] = useState<boolean>(
+    defaultChecked ?? false
```
ShadowOfTheSpace: You already assigned default value to the `defaultChecked` prop

ShadowOfTheSpace (approved):

luiqor (approved):

markgol777 (approved):

PR #2976, opened 2024-12-15
replaced all of the switch elements and adujested the switch component

### LINKED ISSUE #2945: (SP: 2) Replace Switch throughout the project
Replace Switch with the reusable component from a design system

markgol777:
Screenshot 2024-12-15 at 14 31 55

Quiz page

Screenshot 2024-12-15 at 14 32 11

Offers page

Screenshot 2024-12-15 at 14 33 18

Notification tab

### DIFF
--- a/src/components/app-content-switcher/AppContentSwitcher.tsx
+++ b/src/components/app-content-switcher/AppContentSwitcher.tsx
@@ -3,7 +3,7 @@ import { TypographyProps } from '@mui/material/Typography'
 import { Typography } from '@mui/material'
 import { SxProps } from '@mui/system'
 import Tooltip from '@mui/material/Tooltip'
-import Switch from '@mui/material/Switch'
+import AppSwitch from '~/design-system/components/switch/Switch'
 import Stack from '@mui/material/Stack'
 import { defaultStyles } from '~/components/app-content-switcher/AppContentSwitcher.styles'
 import { SwitchContent, SwitchOptions } from '~/types'
@@ -38,12 +38,7 @@ const AppContentSwitcher: FC<SwitcherProps> = ({
   return (
     <Stack sx={{ ...defaultStyles.stack, ...styles } as SxProps}>
       {renderBlock(switchOptions.left, active)}
-      <Switch
-        checked={active}
-        data-testid='switch'
-        onChange={onChange}
-        sx={defaultStyles.switch}
-      />
+      <AppSwitch data-testid='switch' onChange={onChange} />
       {renderBlock(switchOptions.right, !active)}
     </Stack>
   )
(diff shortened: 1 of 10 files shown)


src/components/app-content-switcher/AppContentSwitcher.tsx
```diff
@@ -38,12 +38,7 @@ const AppContentSwitcher: FC<SwitcherProps> = ({
   return (
     <Stack sx={{ ...defaultStyles.stack, ...styles } as SxProps}>
       {renderBlock(switchOptions.left, active)}
-      <Switch
-        checked={active}
-        data-testid='switch'
-        onChange={onChange}
-        sx={defaultStyles.switch}
-      />
+      <AppSwitch data-testid='switch' onChange={onChange} />
```
ShadowOfTheSpace: Why did you remove the `checked` prop here?

src/components/app-content-switcher/AppContentSwitcher.tsx
```diff
@@ -3,7 +3,7 @@ import { TypographyProps } from '@mui/material/Typography'
 import { Typography } from '@mui/material'
 import { SxProps } from '@mui/system'
 import Tooltip from '@mui/material/Tooltip'
-import Switch from '@mui/material/Switch'
+import AppSwitch from '~/design-system/components/switch/Switch'
```
ShadowOfTheSpace: I think we should name all imports without the prefix `App`. As I can see, the `Switch` component has the name `AppSwitch`, but all components inside the design-system folder have names without `App`. So, I think it should be changed in a separate task. What do you think?

ShadowOfTheSpace (changes requested):

src/components/app-content-switcher/AppContentSwitcher.tsx
markgol777: Maybe, but I am not the created of the component
Screenshot 2024-12-15 at 15 01 59
+ we have got a lot of other components which prefix is app so it wouldn't be cool just to go around the whole project and remove one word:)

src/components/app-content-switcher/AppContentSwitcher.tsx
markgol777: Because it was redundant.

PS: Which I found was really weird but it works exactly the same with or without

src/components/app-content-switcher/AppContentSwitcher.tsx
ShadowOfTheSpace: I'm talking about only `design-system` folder here, not about `components` folder. You're importing `Switch` component from `design-system` folder, not from `components`, and all components there (except Switch) have naming without `App`.

markgol777: Fair enough, fixed.

ShadowOfTheSpace (approved):

amoutens (approved):

luiqor (approved):

PR #2990, opened 2024-12-16
Create Menu component

### LINKED ISSUE #2962: (SP: ) Create Menu component aligned with design system
#### Description:
Implement a reusable `Menu` component that meets the design system standards. This component will be used throughout the application to maintain consistency in menu styles, behavior, and accessibility.

Image

#### Requirements:
- styles
  - Follow design system variables for colors, typography, padding, and other styles.
  - Ensure hover, focus, active, and disabled states are defined according to the design system.

- props

Don't forget to cover component with tests and add to the storybook

luiqor:
## Add Menu Component
image

- Disabled / Selected / Hovered as in old design
image

- Simple dropdown menu
image

- Menu with checkboxes
image

-  Nested dropdowns
image

- Removable items (Absent in design, present in prod notification menu)
image

## Add MenuItem Component
- Color variants + align variants
image

- Disabled, Danger color variant
image


src/design-system/components/menu-item/MenuItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+        `s2s-item--density-${density}`,
+        `s2s-item--variant-${variant}`,
+        `s2s-item--color-${colorVariant}`,
+        `s2s-item--align-${alignVariant}`,
+        isToggled && 's2s-item--toggled',
+        isBottomBorder && 's2s-item--bottom-border',
+        isDisabled && 's2s-item--disabled'
+      )}
+      disabled={isDisabled}
+      key={title}
```
ShadowOfTheSpace: You don't need to pass `key` if element is not used in `map`

src/design-system/components/menu/Menu.tsx
```diff
@@ -0,0 +1,181 @@
...
+  defaultOnItemClick,
+  maxHeight,
+  minWidth,
+  noItemsMessage = defaultNoItemsMessage,
+  density = 1,
+  allowToggleMultipleItems = false,
+  isItemsRemovalEnabled = false,
+  ...menuProps
+}: MenuProps) => {
+  const [items, setItems] = useState(menuItems)
```
ShadowOfTheSpace: Add generic type to `useState`

src/design-system/components/menu/Menu.tsx
```diff
@@ -0,0 +1,181 @@
...
+          isToggled={toggledItemsTitles.includes(item.title)}
+          key={item.title}
+          onClick={() => handleItemClick(item)}
+          onRemove={
+            isItemsRemovalEnabled
+              ? () => handleItemRemoval(item.title)
+              : undefined
+          }
+        />,
+        ...(item.nestedMenuItems && toggledItemsTitles.includes(item.title)
```
ShadowOfTheSpace: I would swap those conditions to make them more understandable
```suggestion
        ...(toggledItemsTitles.includes(item.title) && item.nestedMenuItems
```

tests/unit/design-system/components/menu/Menu.spec.jsx
```diff
@@ -0,0 +1,195 @@
...
+
+    render(
+      <div>
+        <button onClick={handleClick}>Open Menu</button>
+        <Menu
+          anchorEl={document.createElement('button')}
+          setAnchorEl={setAnchorEl}
+          menuItems={resourcesMenuItems}
+        />
+      </div>
```
ShadowOfTheSpace:
```suggestion
      <>
        <button onClick={handleClick}>Open Menu</button>
        <Menu
          anchorEl={document.createElement('button')}
          setAnchorEl={setAnchorEl}
          menuItems={resourcesMenuItems}
        />
      </>
```

tests/unit/design-system/components/menu/Menu.spec.jsx
```diff
@@ -0,0 +1,195 @@
...
+
+const resourcesMenuItemsWithIcon = [{ title: 'Lesson', graphics: lessonIcon }]
+
+const noItemsCustomMessage = 'No items available.'
+
+describe('Menu Component', () => {
+  test('should render menu items', () => {
+    render(
+      <Menu
+        anchorEl={document.createElement('div')}
```
ShadowOfTheSpace: Can you replace `document.createElement` with some real element, like you did with `lessonIcon`?

src/design-system/components/menu-item/MenuItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+      const interactiveElement =
+        graphicsRef.current.querySelector<HTMLInputElement>(
+          'input[type="radio"], input[type="checkbox"], [role="radio"], [role="checkbox"]'
+        )
+
+      if (interactiveElement) {
+        interactiveElement.click()
+        return
+      }
+    }
```
ShadowOfTheSpace: Can we get same result with using the <label /> element? Its default behaviour that it is clicking on input that currently inside the label.

ShadowOfTheSpace (changes requested):

src/design-system/components/menu-item/MenuItem.tsx
luiqor: If you mean like that:
```
const handleClick = () => {
    if (graphicsRef.current) {
      graphicsRef.current.click()
      return
    }

    onClick?.()
  }
```
: unfortunately, checkbox then is not marked as selected on click.
 When I am changing the button tag to label here, I can't get rid of onClick triggering two times, even with event.stopPropagation()

tests/unit/design-system/components/menu/Menu.spec.jsx
luiqor: Yeah, used data-testid as I did with lessonIcon, but since it must be rendered, also added logic to render it in beforeEach

src/design-system/components/menu-item/MenuItem.tsx
luiqor: I overlooked it when I was copying from the Menu to a separate component. Thanks for catching that

src/design-system/components/menu-item/MenuItem.tsx
```diff
@@ -0,0 +1,75 @@
+import { FC } from 'react'
```
ShadowOfTheSpace: Please, don't import `FC`, use `React.FC` for typing a component. It's more readable and understandable.
Also, if you are importing just a type - mark this import with `type` (`import { type FC } from 'react'`). It's reducing js bundle size.

src/design-system/components/menu-item/MenuItem.types.ts
```diff
@@ -0,0 +1,15 @@
+import { ReactNode } from 'react'
```
ShadowOfTheSpace:
```suggestion
import { type ReactNode } from 'react'
```

src/design-system/components/menu-item/MenuItem.types.ts
```diff
@@ -0,0 +1,15 @@
...
+import { MenuItemColorVariant } from './MenuItem.constants'
```
ShadowOfTheSpace: Here is a good example, of why you need to import something as a `type` if you use it just as a type.
`MenuItemColorVariant` enum will become a self-calling function when ts will be converted to `js` (check [this](https://www.typescriptlang.org/play/?verbatimModuleSyntax=false#code/KYDwDg9gTgLgBMAdgVwLZwLJOQSRsVANQEMoBLYxeAbwFgAoOOAEWADNjkAbeAXjgDkAE3aceAgDQMmAOWABnfELj8BiBUoEMAvkA) out). So if you import this thing without a `type` prefix, this will be actually imported into this file in `js`. But we have no types inside `js`, so it will be an unused import here, and the rest of the file will be empty.
If you will add a type prefix (because you are not using any value of that enum here), `ts` will drop this import, and we will have this file empty at all.
It is very useful when you import something from a third-party library because it really helps to reduce the `js bundle size`.

src/design-system/components/menu/Menu.tsx
```diff
@@ -0,0 +1,196 @@
+import { FC, useState } from 'react'
```
ShadowOfTheSpace: Same here about `FC`

src/design-system/components/menu/Menu.tsx
```diff
@@ -0,0 +1,196 @@
...
+  const setToggledItemsTitles = (
+    updater: (prevItems: string[]) => string[]
+  ) => {
+    onToggleItemsChange
+      ? onToggleItemsChange(updater(toggledItemsTitles))
+      : setInternalToggledItemsTitles(updater)
+  }
+
+  const toggleAsSingleItem = (itemTitle: string) => {
+    setToggledItemsTitles((previousItems) =>
```
ShadowOfTheSpace: If the arrow function doesn't fit into a single line - add curly braces. It will be a lot more readable with them.

src/design-system/components/menu/Menu.tsx
```diff
@@ -0,0 +1,196 @@
...
+      previousItems.includes(itemTitle) ? [] : [itemTitle]
+    )
+  }
+
+  const toggleAsOneOfMultipleItems = (itemTitle: string) => {
+    setToggledItemsTitles((previousItems) =>
```
ShadowOfTheSpace: Same here about arrow function

ShadowOfTheSpace (changes requested):

src/design-system/components/menu-item/MenuItem.types.ts
luiqor: Thank you for the great explanation! 🚀

markgol777 (approved):

Mike-Popovych (approved):

src/design-system/components/menu-item/MenuItem.tsx
```diff
@@ -0,0 +1,74 @@
+import { MenuItem as MuiMenuItem } from '@mui/material'
+import { ArrowDropDown, ArrowDropUp, CloseRounded } from '@mui/icons-material'
+
+import { MenuItemProps as CommonMenuItemProps } from './MenuItem.types'
+import { MenuItemColorVariant, MenuItemVariant } from './MenuItem.constants'
+import { cn } from '~/utils/cn'
+
+import '~scss-components/menu-item/MenuItem.scss'
```
ShadowOfTheSpace: I think you can leave here `./MenuItem.scss` as it is already in one folder

src/design-system/components/menu-item/MenuItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+  title,
+  additionalInfo,
+  alignVariant = 'left',
+  colorVariant = MenuItemColorVariant.Default,
+  density = 1,
+  graphics,
+  isDropdown = false,
+  isToggled = false,
+  isBottomBorder = false,
+  isDisabled = false,
```
ShadowOfTheSpace: Check if you really need to have defined values (`false` as default), probably falsy values enough here

src/design-system/components/menu/Menu.tsx
```diff
@@ -0,0 +1,196 @@
...
+  MenuItemProps as CommonMenuItemProps,
+  OnItemClickArgs
+} from '~/design-system/components/menu-item/MenuItem.types'
+import MenuItem from '../menu-item/MenuItem'
+import {
+  MenuItemColorVariant,
+  MenuItemVariant
+} from '../menu-item/MenuItem.constants'
+
+import '~scss-components/menu/Menu.scss'
```
ShadowOfTheSpace: Probably same here about import

src/design-system/components/menu/Menu.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+const removeAllItemsTitle = 'Clear all'
+const defaultNoItemsMessage = 'No items.'
```
ShadowOfTheSpace: Is it needs to be translated?

src/design-system/components/menu-item/MenuItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+interface MenuItemProps extends CommonMenuItemProps {
+  isDropdown?: boolean
+  density?: 1 | 2
+  isToggled?: boolean
+  onRemove?: () => void
+  variant?: MenuItemVariant
+}
+
+const MenuItem: React.FC<MenuItemProps> = ({
```
ShadowOfTheSpace: Please, use `forwardRef` here

src/design-system/components/menu/Menu.tsx
```diff
@@ -0,0 +1,196 @@
...
+  isItemsRemovalEnabled?: boolean
+  noItemsMessage?: string
+  maxHeight?: number
+  minWidth?: number
+  transformOrigin?: PopoverOrigin
+  toggledItemsTitles?: string[]
+  onToggleItemsChange?: (newTitles: string[]) => void
+}
+
+const Menu: React.FC<MenuProps> = ({
```
ShadowOfTheSpace: Please, use `forwardRef` here too

ShadowOfTheSpace (changes requested):

src/design-system/components/menu/Menu.tsx
luiqor: I guess there might be such an option. So I added an optional removeAllItemsTitle prop to the Menu component, alongside the existing noItemsMessage prop, making it possible to use custom translations directly in the frontend app, instead of relying on the design system

ShadowOfTheSpace (approved):

PR #2992, opened 2024-12-17
Created a wrapper over `useQuery` hook from react-query

### LINKED ISSUE #2991: Create a custom wrapper for `useQuery` hook from react-query for future replacement
**Description:**
Develop a custom wrapper around the useQuery hook provided by react-query. The goal is to centralize configurations and streamline the eventual replacement of the underlying data-fetching library if needed. This custom hook should:

- Handle default options (e.g., caching, retries).
- Provide a consistent interface across the project.
- Allow for custom transformations or pre/post-processing of data.
- Facilitate easy debugging and logging for query operations.

**Context:**
This approach decouples the application code from the specific implementation of react-query, enabling future flexibility and reducing the impact of potential library changes.

Made1ra:
**Description:**
Developed a custom wrapper around the useQuery hook provided by react-query. The goal is to centralize configurations and streamline the eventual replacement of the underlying data-fetching library if needed. This custom hook should:

- Handle default options (e.g., caching, retries).
- Provide a consistent interface across the project.
- Allow for custom transformations or pre/post-processing of data.
- Facilitate easy debugging and logging for query operations.

**Context:**
This approach decouples the application code from the specific implementation of react-query, enabling future flexibility and reducing the impact of potential library changes.


markgol777 (approved):

src/QueryProvider.tsx
```diff
@@ -0,0 +1,20 @@
+import { FC, ReactNode } from 'react'
```
ShadowOfTheSpace: Please, do not import FC, it's unreadable to have `FC` as a type. If you see `React.FC` you know that this is a react function component, but when you see just `FC` as a type, you need to check, where this `FC` came from, especially if you are new in this codebase.

Also, if you are importing types, make sure to add `type` prefix to this imports. It is a very good practice, especially if you are using non-type values as a type. This approuch reduces js bundle size.

So, your import will be like this: `import { type ReactNode } from 'react'

src/hooks/use-query.tsx
```diff
@@ -0,0 +1,41 @@
...
+  TQueryFnData,
+  TError = AxiosError,
+  TData = TQueryFnData,
+  TQueryKey extends unknown[] = []
+>({
+  queryKey,
+  queryFn,
+  options
+}: UseQueryProps<TQueryFnData, TError, TData, TQueryKey>) => {
+  const { isLoading, error, data, refetch } = useReactQuery<
```
ShadowOfTheSpace: There are more fields in `reactQuery` hook, why did you choose only these?

src/hooks/use-query.tsx
```diff
@@ -0,0 +1,41 @@
+import {
+  useQuery as useReactQuery,
+  UseQueryOptions
+} from '@tanstack/react-query'
+import { AxiosError } from 'axios'
```
ShadowOfTheSpace:
```suggestion
import { type AxiosError } from 'axios'
```

src/hooks/use-query.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+interface UseQueryProps<
+  TQueryFnData,
+  TError,
+  TData,
+  TQueryKey extends unknown[]
+> {
+  queryKey: TQueryKey
+  queryFn: () => Promise<TQueryFnData>
```
ShadowOfTheSpace: I don't think that `queryFn` should return Promise with type of `QueryFnData` (or it's very confusing). Try to add better type names, that will be understandable without going to tanstack docs)
https://github.com/TanStack/query/discussions/1477

src/hooks/use-query.tsx
```diff
@@ -0,0 +1,41 @@
...
+  options?: UseQueryOptions<TQueryFnData, TError, TData>
+}
+
+const useQuery = <
+  TQueryFnData,
+  TError = AxiosError,
```
ShadowOfTheSpace: I don't think that we need all those fields inside `AxiosError` object. If we will use this error inside our components, we will have to manually map this object to type that we need and take only those fields, that we need. This logic will be duplicated inside all components that are using the error object.
Maybe it would be better to extract only needed fields here (with some helper/mapper function ofc)? 
image

ShadowOfTheSpace (changes requested):

ShadowOfTheSpace (approved):

src/hooks/use-query.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface UseQueryProps<TData, TError, TSelect> {
+  queryKey: QueryKey
+  queryFn: () => Promise<TData>
+  options?: UseQueryOptions<TData, TError, TSelect>
+}
+
+const useQuery = <TData, TError = ErrorResponse, TSelect = TData>({
+  queryKey,
+  queryFn,
```
dudchakk, line 16: does `queryFn` stand for queryFunction? If that, please rename it to `queryFunction`, as it is more readable

Made1ra: `queryFn` is a name from `@tanstack/react-query` library, I think this name is better here for consistency with [@tanstack/react-query API](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery)

dudchakk: okay, got it, then it’s better to leave it like that

dudchakk (approved):

PR #3007, opened 2024-12-20
Fixed type incompatibility for size prop of AppButton in OfferActions Component

### LINKED ISSUE #2979: (SP: 1) Fix type incompatibility for size prop of AppButton in OfferActions Component
There is a type mismatch error in the OfferActions component where the `size` prop passed to `AppButton` is incompatible with the expected type in `AppButtonProps`.

File Path:
`src/components/offer-card/offer-actions/OfferActions.tsx`

Code:
image

Problem:
image

SofiiaYevush:
Extended the `size` prop type for the `AppButton` component to include the `SizeEnum` values (`'small' | 'medium' | 'large' | 'extraLarge' | 'xxl' `). This enables passing `size` values as strings, improving compatibility with MUI.

This change is aligned with SizeEnum definitions:
`common.enums.ts`:
image

### DIFF
--- a/src/components/app-button/AppButton.tsx
+++ b/src/components/app-button/AppButton.tsx
@@ -12,7 +12,7 @@ interface AppButtonProps extends Omit<ButtonProps, 'size'> {
   disabled?: boolean
   component?: ElementType
   to?: string
-  size?: 'small' | 'medium' | 'large' | 'extraLarge' | 'xxl' | SizeEnum | null
+  size?: `${SizeEnum}` | SizeEnum | null
 }
 
 const AppButton: FC<AppButtonProps> = ({
--- a/src/components/app-button/AppButton.tsx
+++ b/src/components/app-button/AppButton.tsx
@@ -12,7 +12,7 @@ interface AppButtonProps extends Omit<ButtonProps, 'size'> {
   disabled?: boolean
   component?: ElementType
   to?: string
-  size?: SizeEnum | null
+  size?: 'small' | 'medium' | 'large' | 'extraLarge' | 'xxl' | SizeEnum | null
 }
 
 const AppButton: FC<AppButtonProps> = ({


markgol777 (approved):

src/components/app-button/AppButton.tsx
```diff
@@ -12,7 +12,7 @@ interface AppButtonProps extends Omit<ButtonProps, 'size'> {
   disabled?: boolean
   component?: ElementType
   to?: string
-  size?: SizeEnum | null
+  size?: 'small' | 'medium' | 'large' | 'extraLarge' | 'xxl' | SizeEnum | null
```
ShadowOfTheSpace: You can do it in a cleaner way
```suggestion
  size?: `${SizeEnum}` | SizeEnum | null
```

ShadowOfTheSpace (changes requested):

src/components/app-button/AppButton.tsx
ShadowOfTheSpace: And also, why did you estimate this issue for 2 SP? Did you have some problems finding out what caused this error?

SofiiaYevush: > And also, why did you estimate this issue for 2 SP? Did you have some problems finding out what caused this error?

I tried several other ways to fix the error so as not to change the `AppButton` on which many other files depended, but none of them were good. So I defined 2 SP as it took a while to find the best solution that would not cause errors in other files. Do you suggest estimating less?

SofiiaYevush: > You can do it in a cleaner way

Done.

ShadowOfTheSpace: > > And also, why did you estimate this issue for 2 SP? Did you have some problems finding out what caused this error?
> 
> I tried several other ways to fix the error so as not to change the `AppButton` on which many other files depended, but none of them were good. So I defined 2 SP as it took a while to find the best solution that would not cause errors in other files. Do you suggest estimating less?

Actually, estimation of the bug should be done by you, and only you. But in my opinion, 2 SP is a little bit too much for a one line fix. Especially when we will replace AppButton with Button from design-system folder, so this fix is temporary. But, as I said before - it's up to you.

ShadowOfTheSpace (approved):

luiqor (approved):

PR #3024, opened 2024-12-24
replaced useAxios hook with useQuery and fixed all the bugs with it

markgol777:
Replaced useAxios with useQuery in PopularCategories component

image

### DIFF
--- a/src/components/popular-categories/PopularCategories.tsx
+++ b/src/components/popular-categories/PopularCategories.tsx
@@ -1,6 +1,6 @@
 import Box from '@mui/material/Box'
 import { SxProps } from '@mui/material/styles'
-import { FC, useCallback, useMemo } from 'react'
+import { FC, useMemo, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useNavigate } from 'react-router-dom'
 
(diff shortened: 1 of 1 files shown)


src/components/popular-categories/PopularCategories.tsx
```diff
@@ -24,60 +23,66 @@ import {
...
+    return response.data
+  }, [itemsToShow])
+
+  const { data, isLoading } = useQuery<ItemsWithCount<CategoryInterface>>({
+    queryKey: ['popularCategories', itemsToShow],
+    queryFn: getCategories
   })
 
-  const oppositeRole = getOpositeRole(userRole)
+  const oppositeRole = useMemo(() => getOpositeRole(userRole), [userRole])
```
ShadowOfTheSpace, line 65: You need to use the `useMemo` hook for some complex calculations that need to be stable across rerenders. `getOpositeRole` is just a ternary operator, so `useMemo` is too much for that. Overmemoization can be worth more than no memoization at all.

src/components/popular-categories/PopularCategories.tsx
```diff
@@ -1,21 +1,20 @@
 import Box from '@mui/material/Box'
 import { SxProps } from '@mui/material/styles'
-import { FC, useCallback, useMemo } from 'react'
+import { FC, useMemo, useCallback } from 'react'
```
ShadowOfTheSpace: Do not import `FC`, use `React.FC`, please

src/components/popular-categories/PopularCategories.tsx
```diff
@@ -1,21 +1,20 @@
...
 import CardWithLink from '~/components/card-with-link/CardWithLink'
 import CardsList from '~/components/cards-list/CardsList'
 import Loader from '~/components/loader/Loader'
 import TitleWithDescription from '~/components/title-with-description/TitleWithDescription'
-import useAxios from '~/hooks/use-axios'
 import useBreakpoints from '~/hooks/use-breakpoints'
 import { useAppSelector } from '~/hooks/use-redux'
 import { authRoutes } from '~/router/constants/authRoutes'
 import { categoryService } from '~/services/category-service'
+import { AxiosResponse } from 'axios'
```
ShadowOfTheSpace:
```suggestion
import { type AxiosResponse } from 'axios'
```

ShadowOfTheSpace (changes requested):

src/components/popular-categories/PopularCategories.tsx
markgol777: You're right but since it was there before I started working on the component I just let it be

src/components/popular-categories/PopularCategories.tsx
markgol777: sure

Made1ra (approved):

markgol777:
Outdated

PR #3040, opened 2024-12-30
Enhance useQuery hook

### LINKED ISSUE #3039: (SP: 5) Enhance useQuery hook
- [x] Fix types mismatches
- [x] Define options and properties that will be used in project

ShadowOfTheSpace:
### `useQuery` hook can be used with the next properties:
- `queryFn` - async function for querying data (required)
- `queryKey` - an array of unique keys for each `useQuery` call (required)
- `options` - additional properties for query settings (optional)
    - `enabled` - is auto fetching enabled (can be derived from falsy or truthy value)
    - `initialData` - initial value of the query result (should be used if `data` can't be `undefined`)
    - `select` - function for modifying `data` object structure (should be used if data needs to be transformed to another shape)
    - `staleTime` - number of times that query result should be **valid** (if `Infinity` then data is always valid, and fetching will not run). **Do not use with `initialData` because it will maintain `initialData` as a result of the query forever**.
    
### `useQuery` hook returns the next parameters:
- `data` - a result of the query (`undefined` if no `initialData` was provided)
- `error` - an error that was catched by hook (currently it's `ResponseError` class)
- `isError` - `boolean` representation of `error`
- `isFetching` - returns `true` on each run of `queryFn`
- `isLoading` - returns `true` only on the first run of `queryFn`
- `refetch` - function that triggers refetching (if `enabled` is `false` then only `refetch` call can run fetching)

### `data` type behaviour:
- when the `options` prop isn't provided:

  image_2024-12-30_12-49-21

- when provided an empty `options` prop:

  image_2024-12-30_12-49-53

- when the `select` function is provided:

  image_2024-12-30_12-50-44

- when the `initialData` prop is provided:

  image_2024-12-30_12-50-11

- when the `select` and `initialData` are provided:

  image_2024-12-30_12-52-36

Also created `ResponseError` class that will be used in base service for each HTTP request for consistency:
```typescript
export class ResponseError extends Error {
  code?: string
  status?: number

  constructor({ code, message, status }: Partial<ErrorResponse>) {
    super(message)

    this.code = code
    this.status = status
  }
}
```

### DIFF
--- a/src/exceptions/response-error.ts
+++ b/src/exceptions/response-error.ts
@@ -1,0 +1,15 @@
+import { type ErrorResponse } from '~/types'
+
+class ResponseError extends Error {
+  code?: string
+  status?: number
+
+  constructor({ code, message, status }: Partial<ErrorResponse>) {
+    super(message)
+
+    this.code = code
+    this.status = status
+  }
+}
+
+export { ResponseError }
--- a/src/hooks/use-query.tsx
+++ b/src/hooks/use-query.tsx
@@ -8,7 +8,7 @@ import {
   useQuery as useReactQuery
 } from '@tanstack/react-query'
 
-import { type ResponseError } from '~/types'
+import { type ResponseError } from '~/exceptions'
 
 type RequiredQueryOptions = 'queryFn' | 'queryKey'
 type OptionalQueryOptions = 'enabled' | 'initialData' | 'select' | 'staleTime'
--- a/src/hooks/use-query.tsx
+++ b/src/hooks/use-query.tsx
@@ -1,28 +1,104 @@
 import {
-  useQuery as useReactQuery,
-  QueryKey,
-  UseQueryOptions
+  type DefinedInitialDataOptions,
+  type DefinedUseQueryResult,
+  type QueryKey,
+  type UndefinedInitialDataOptions,
+  type UseQueryOptions,
+  type UseQueryResult,
+  useQuery as useReactQuery
 } from '@tanstack/react-query'
-import { ErrorResponse } from '~/types'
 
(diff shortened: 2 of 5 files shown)


src/exceptions/response-error.ts
```diff
@@ -0,0 +1,15 @@
+import { type ErrorResponse } from '~/types'
+
+class ResponseError extends Error {
+  code?: string
+  status?: number
```
dudchakk, line 5: why these fields are optional?

Made1ra (approved):

src/hooks/use-query.tsx
```diff
@@ -1,28 +1,104 @@
...
+  TData = TQueryFnData,
+  TQueryKey extends QueryKey = QueryKey
+>(
+  properties: {
+    options: Pick<
+      DefinedInitialDataOptions<TQueryFnData, TError, TData, TQueryKey>,
+      OptionalQueryOptions
+    >
+  } & RequiredQueryProperties<TQueryFnData, TError, TData, TQueryKey>
+): Pick<DefinedUseQueryResult<TData, TError>, ReturningQueryParams>
```
dudchakk: maybe you should move the long type definitions into separate types, so that the useQuery definition won't be as accumulated

dudchakk (commented):
amazing job!
just add the story points to the issue :)

src/exceptions/response-error.ts
ShadowOfTheSpace: If we will have error not from our server, we won't know status and code (it will be internal error of Axios or network)

dudchakk (approved):

Mike-Popovych (approved):

PR #3043, opened 2024-12-30
Create base service for all HTTP requests

### LINKED ISSUE #3042: (SP: 3) Create base service for all HTTP requests
The service needs to be updated with the following features::

- [x] response handling: extract and handle `response.data` consistently for all requests
- [x] implement strict type-checking
- [x] сonvert all errors into a standardized `ResponseError` exception for consistent error management

ShadowOfTheSpace:
`baseService` contains only one function - `request` with the following parameters:
 - `method` - can be one of: `GET`, `POST`, `PATCH` or `DELETE` (required)
 - `url` - request url with dynamic and query parameters (required)
 - `data` - body of the request (optional)
 - `headers` - headers of the request (optional)
 - `timeout` - max time of the request in ms (optional)
 
`request` function has one generic parameter (`unknown` by default) which is used to type the response body. 
**It can't be inferred from the statement, so you should always provide a type that represents the data of the response**.

If an error is thrown inside the `baseService`, the error will be transformed into a `ResponseError` exception.

For `baseService` usage, put `baseService.request` function call inside a particular service instead of `axiosClient`, for example:
```typescript
import { URLs } from '~/constants/request'
import { type Country } from '~/types'
import { createUrlPath } from '~/utils/helper-functions'
import { baseService } from './base-service'

export const LocationService = {
  getCountries: () => {
    return baseService.request<Country[]>({
      method: 'GET',
      url: URLs.location.getCountries
    })
  },
  getCities: (country: string) => {
    return baseService.request<string[]>({
      method: 'GET',
      url: createUrlPath(URLs.location.getCities, country)
    })
  }
}

```

### DIFF
--- a/src/services/base-service.ts
+++ b/src/services/base-service.ts
@@ -1,21 +1,31 @@
-import { isAxiosError } from 'axios'
+import { isAxiosError, type RawAxiosRequestHeaders } from 'axios'
 import { ResponseError } from '~/exceptions'
 import { axiosClient } from '~/plugins/axiosClient'
 import { type ErrorResponse, type HttpMethod } from '~/types'
 
 type RequestParams = {
   data?: unknown
+  headers?: RawAxiosRequestHeaders
   method: HttpMethod
+  timeout?: number
   url: string
 }
 
 export const baseService = {
-  request: async <T = unknown>({ method, url, data }: RequestParams) => {
+  request: async <T = unknown>({
+    data,
+    headers,
+    method,
+    timeout,
+    url
+  }: RequestParams) => {
     try {
       const response = await axiosClient.request<T>({
-        url,
+        data,
+        headers,
         method,
-        data
+        timeout,
+        url
       })
 
       return response.data
--- a/src/services/base-service.ts
+++ b/src/services/base-service.ts
@@ -1,0 +1,35 @@
+import { isAxiosError } from 'axios'
+import { ResponseError } from '~/exceptions'
+import { axiosClient } from '~/plugins/axiosClient'
+import { type ErrorResponse, type HttpMethod } from '~/types'
+
+type RequestParams = {
+  data?: unknown
+  method: HttpMethod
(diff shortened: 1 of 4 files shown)


nebby2105 (approved):

Made1ra (approved):

src/services/base-service.ts
markgol777, line 1: Consider adding headers (we use it in some places of our projects) and timeout for longer requests

markgol777 (changes requested):

src/services/base-service.ts
markgol777:
```js
import { isAxiosError } from 'axios';
import { ResponseError } from '~/exceptions';
import { axiosClient } from '~/plugins/axiosClient';
import { type ErrorResponse, type HttpMethod } from '~/types';

type RequestParams = {
  data?: unknown;
  method: HttpMethod;
  url: string;
  headers?: Record<string, string>; // Allow custom headers
  timeout?: number; // Optional timeout
};

const handleError = (error: unknown): never => {
  if (isAxiosError(error) && error.response) {
    const serverError = error.response.data as ErrorResponse;
    throw new ResponseError(serverError);
  }

  throw new ResponseError({
    code: 'UNKNOWN_ERROR',
    message: 'An unknown error occurred. Please try again later.',
  });
};

export const baseService = {
  request: async <T = unknown>({
    method,
    url,
    data = {},
    headers = {},
    timeout = 5000, // Default timeout of 5 seconds
  }: RequestParams): Promise<T> => {
    try {
      const response = await axiosClient.request<T>({
        url,
        method,
        data,
        headers,
        timeout,
      });

      return response.data;
    } catch (error) {
      handleError(error);
    }
  },
};

(something like this)
```

ShadowOfTheSpace: Thanks for suggestion!

markgol777 (approved):

PR #3045, opened 2024-12-31
Replace useAxios with useQuery on AddResources

### LINKED ISSUE #3015: Replace `useAxios` with `useQuery` on the AddResources.tsx
Description

Replace the `useAxios` hook with a custom useQuery hook on the My Cooperations page to improve data fetching, caching, and error handling.

Tasks

- [ ] Identify where `useAxios` is used
- [ ] Replaces all of the instances of `useAxios` with useQuery
- [ ] Update tests if necessary
- [ ] Ensure that everything works as intended

nebby2105:
#3015 
- [x] Replaced all instances of `useAxios` with `useQuery` in AddResources.
- [x] Updated tests to accommodate the new implementation.
- [x] Verified that the data fetching and error handling work as intended with useQuery.

Created new service functions (`getUsersLessonsQuery`, `getQuizzesQuery`, `getAttachmentsQuery`, `getQuestionsQuery`) that utilize the `baseService` for API calls instead of directly using `axiosClient`.

The existing service functions (`getUsersLessons`, `getQuizzes`, `getAttachments`, `getQuestions`) remain unmodified to ensure backward compatibility and uninterrupted functionality of components relying on them.
These new functions are designed to replace the old ones once all relevant components transition to using useAxios.

### DIFF
--- a/src/containers/add-resources/AddResources.tsx
+++ b/src/containers/add-resources/AddResources.tsx
@@ -3,7 +3,8 @@ import { useCallback, useState } from 'react'
 import { useAppDispatch } from '~/hooks/use-redux'
 import useSelect from '~/hooks/table/use-select'
 import useSort from '~/hooks/table/use-sort'
-import useAxios from '~/hooks/use-axios'
+// import useAxios from '~/hooks/use-axios'
+import useQuery from '~/hooks/use-query'
 import useBreakpoints from '~/hooks/use-breakpoints'
 
 import { useModalContext } from '~/context/modal-context'
@@ -82,10 +83,26 @@ const AddResources = <T extends CourseResource | Question>({
     [dispatch]
   )
 
-  const { loading, response } = useAxios<ItemsWithCount<T>>({
-    service: getMyResources,
-    defaultResponse: defaultResponses.itemsWithCount,
-    onResponseError
+  // const { loading, response } = useAxios<ItemsWithCount<T>>({
+  //   service: getMyResources,
+  //   defaultResponse: defaultResponses.itemsWithCount,
+  //   onResponseError
+  // })
+
+  const { isLoading: loading, data } = useQuery<
+    ItemsWithCount<T>,
+    ErrorResponse
+  >({
+    queryKey: ['resources'],
+    queryFn: async () => {
+      try {
+        const response = await getMyResources()
+        return response.data
+      } catch (error) {
+        onResponseError(error as ErrorResponse)
+        return defaultResponses.itemsWithCount
+      }
+    }
   })
 
   const onRowClick = useCallback(
@@ -123,7 +140,8 @@ const AddResources = <T extends CourseResource | Question>({
 
   const getItems = useCallback(
     (inputValue: string, selectedCategories: string[]) => {
-      return response.items.filter((item) => {
+      if (!data?.items) return []
+      return data?.items.filter((item) => {
         const titleMatch =
           'title' in item
             ? item.title
@@ -146,7 +164,7 @@ const AddResources = <T extends CourseResource | Question>({
         return titleMatch && categoryMatch
       })
     },
-    [response.items]
+    [data?.items]
   )
 
   const props = {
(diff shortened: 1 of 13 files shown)


src/containers/add-resources/AddResources.tsx
```diff
@@ -70,23 +69,24 @@ const AddResources = <T extends CourseResource | Question>({
...
-    (error?: ErrorResponse) => {
+  const { data, isLoading, error } = useQuery({
+    queryKey: ['resources', sort, resourceTab],
+    queryFn: getMyResources,
+    options: {
+      initialData: defaultResponses.itemsWithCount
+    }
+  })
+
+  useEffect(() => {
```
Made1ra, line 80: Try to remove the `useEffect`. I guess it should work without it.
```suggestion
```

src/services/resource-service.ts
```diff
@@ -25,13 +25,24 @@ import {
...
     params?: GetResourcesParams
   ): Promise<AxiosResponse<ItemsWithCount<Lesson>>> =>
     await axiosClient.get(URLs.resources.lessons.get, { params }),
+
+  getUsersLessonsQuery: (params?: GetResourcesParams) => {
+    return baseService.request<ItemsWithCount<Lesson>>({
+      method: 'GET',
+      url: getFullUrl({
+        pathname: URLs.resources.lessons.get,
+        searchParameters: params
```
Made1ra, line 42: Are you sure that it's working without destructuring?
```suggestion
        searchParameters: { ...params }
```

src/services/resource-service.ts
```diff
@@ -47,6 +58,16 @@ export const ResourceService = {
     params?: GetResourcesParams
   ): Promise<AxiosResponse<ItemsWithCount<Quiz>>> =>
     await axiosClient.get(URLs.quizzes.get, { params }),
+
+  getQuizzesQuery: (params?: GetResourcesParams) => {
+    return baseService.request<ItemsWithCount<Quiz>>({
+      method: 'GET',
+      url: getFullUrl({
+        pathname: URLs.quizzes.get,
+        searchParameters: params
```
Made1ra, line 67: Are you sure that it's working without destructuring?
```suggestion
        searchParameters: { ...params }
```

src/services/resource-service.ts
```diff
@@ -62,6 +83,16 @@ export const ResourceService = {
     params?: GetResourcesParams
   ): Promise<AxiosResponse<ItemsWithCount<Attachment>>> =>
     await axiosClient.get(URLs.resources.attachments.get, { params }),
+
+  getAttachmentsQuery: (params?: GetResourcesParams) => {
+    return baseService.request<ItemsWithCount<Attachment>>({
+      method: 'GET',
+      url: getFullUrl({
+        pathname: URLs.resources.attachments.get,
+        searchParameters: params
```
Made1ra, line 92: Are you sure that it's working without destructuring?
```suggestion
        searchParameters: { ...params }
```

src/services/resource-service.ts
```diff
@@ -81,6 +112,16 @@ export const ResourceService = {
   ): Promise<AxiosResponse<ItemsWithCount<Question>>> => {
     return axiosClient.get(URLs.resources.questions.get, { params })
   },
+
+  getQuestionsQuery: (params?: GetResourcesParams) => {
+    return baseService.request<ItemsWithCount<Question>>({
+      method: 'GET',
+      url: getFullUrl({
+        pathname: URLs.resources.questions.get,
+        searchParameters: params
```
Made1ra, line 121: Are you sure that it's working without destructuring?
```suggestion
        searchParameters: { ...params }
```

Made1ra (changes requested):
How many times are changed services used in other files? I really don't like the idea of creating temporary services. Maybe it's better to replace `useAxios` in related files if there are only a few of them.

src/containers/add-resources/AddResources.tsx
nebby2105: I think it's more logical to show an alert only when an error occurs, not on every render. I removed the `useEffect` as you suggested and placed a `console.log` inside the `if (error)` statement. However, in this setup, the function is called twice for every error, whereas with the `useEffect`, it was called only once.

Made1ra: I think it's called twice because you are doing it in development with strict mode on. Try to use the production build to test it.

Made1ra (approved):

SofiiaYevush (approved):

ShadowOfTheSpace (approved):

PR #3048, opened 2025-01-01
Replaced useAxios with useQuery in popular categories

### LINKED ISSUE #3011: Replace `useAxios` with `useQuery` on the Popular categories page
Description

Replace the useAxios hook with a custom useQuery hook on the My Cooperations page to improve data fetching, caching, and error handling.

Tasks

- [x] Identify where useAxios is used
- [x] Replaces all of the instances of useAxios with useQuery
- [x] Update tests if necessary
- [x] Ensure that everything works as intended

markgol777:
image

Took a little longer as we ran into technicalities but we figured it out eventually

Replaced useAxios with useQuery

image

### DIFF
--- a/src/services/category-service.ts
+++ b/src/services/category-service.ts
@@ -13,11 +13,9 @@ import { getFullUrl } from '~/utils/helper-functions'
 
 export const categoryService = {
   getCategories: (params?: Partial<CategoriesParams>) => {
-    const { sort, ...rest } = params || {}
-
     const resultUrl = getFullUrl({
       pathname: URLs.categories.get,
-      searchParameters: { ...sort, ...rest }
+      searchParameters: params
     })
 
     return baseService.request<ItemsWithCount<CategoryInterface>>({
--- a/src/services/category-service.ts
+++ b/src/services/category-service.ts
@@ -19,7 +19,7 @@ export const categoryService = {
       pathname: URLs.categories.get,
       searchParameters: { ...sort, ...rest }
     })
-    console.log(resultUrl)
+
     return baseService.request<ItemsWithCount<CategoryInterface>>({
       method: 'GET',
       url: resultUrl
--- a/src/services/category-service.ts
+++ b/src/services/category-service.ts
@@ -8,12 +8,22 @@ import {
   CategoriesParams,
   ItemsWithCount
 } from '~/types'
+import { baseService } from '~/services/base-service'
+import { getFullUrl } from '~/utils/helper-functions'
 
 export const categoryService = {
-  getCategories: (
-    params?: Partial<CategoriesParams>
-  ): Promise<AxiosResponse<ItemsWithCount<CategoryInterface>>> => {
-    return axiosClient.get(URLs.categories.get, { params })
+  getCategories: (params?: Partial<CategoriesParams>) => {
+    const { sort, ...rest } = params || {}
+
+    const resultUrl = getFullUrl({
+      pathname: URLs.categories.get,
+      searchParameters: { ...sort, ...rest }
+    })
+    console.log(resultUrl)
+    return baseService.request<ItemsWithCount<CategoryInterface>>({
+      method: 'GET',
+      url: resultUrl
+    })
   },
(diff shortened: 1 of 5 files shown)


src/services/category-service.ts
```diff
@@ -8,12 +8,22 @@ import {
...
-  ): Promise<AxiosResponse<ItemsWithCount<CategoryInterface>>> => {
-    return axiosClient.get(URLs.categories.get, { params })
+  getCategories: (params?: Partial<CategoriesParams>) => {
+    const { sort, ...rest } = params || {}
+
+    const resultUrl = getFullUrl({
+      pathname: URLs.categories.get,
+      searchParameters: { ...sort, ...rest }
+    })
+    console.log(resultUrl)
```
Made1ra:
```suggestion
```

Made1ra (changes requested):

src/services/category-service.ts
markgol777: Fixed

Made1ra (approved):

src/services/category-service.ts
```diff
@@ -8,12 +8,22 @@ import {
...
-    params?: Partial<CategoriesParams>
-  ): Promise<AxiosResponse<ItemsWithCount<CategoryInterface>>> => {
-    return axiosClient.get(URLs.categories.get, { params })
+  getCategories: (params?: Partial<CategoriesParams>) => {
+    const { sort, ...rest } = params || {}
+
+    const resultUrl = getFullUrl({
+      pathname: URLs.categories.get,
+      searchParameters: { ...sort, ...rest }
+    })
```
ShadowOfTheSpace: `getFullUrl` is compatible with nested objects like `sort`
If you will have a typescript error like `Index signature for type 'string' is missing in type`, just change the definition `CategoriesParams` from `interface` to a `type`.

```suggestion
  getCategories: (params?: Partial<CategoriesParams>) => {
    const resultUrl = getFullUrl({
      pathname: URLs.categories.get,
      searchParameters: params
    })
```

tests/unit/components/popular-categories/PopularCategories.spec.jsx
```diff
@@ -54,12 +55,6 @@ describe('PopularCategories', () => {
     expect(title).toBeInTheDocument()
   })
 
-  it('render card correctly', async () => {
-    const card = await screen.findByText(t(`categories.${titleToCamel('Math')}`, { defaultValue: 'Math' }))
-
-    expect(card).toBeInTheDocument()
-  })
-
```
ShadowOfTheSpace, line 62: Why did you remove this test?

ShadowOfTheSpace (changes requested):

tests/unit/components/popular-categories/PopularCategories.spec.jsx
markgol777: Honestly I couldn't get it to work but the thing itself works fine..

src/services/category-service.ts
markgol777: That didn't really do the trick it still giving me the error

tests/unit/components/popular-categories/PopularCategories.spec.jsx
```diff
@@ -38,6 +38,7 @@ const title = 'common.popularCategories'
 const description = 'studentHomePage.popularCategories.description'
 
 describe('PopularCategories', () => {
+
   beforeEach(async () => {
     await waitFor(() => {
       mockAxiosClient.onGet(URLs.categories.get).reply(200, mockResponse)
```
ShadowOfTheSpace, line 44:
```suggestion
      mockAxiosClient.onGet(new RegExp(URLs.categories.get)).reply(200, mockResponse)
```
Try to add `RegExp` for URL mapping on GET request. Probably it's because before `baseService` we have to put searchParams inside the `axiosClient` call, but right now we generate the resulting URL with search params with another approach (`getFullUrl`).

ShadowOfTheSpace (approved):

Mike-Popovych (approved):

PR #3156, opened 2025-01-29
Fixed Categories page not found error

### LINKED ISSUE #3137: (SP: ) Fix Categories Page not found
Actual result: Categories Page doesn't work.

Image

Image

Make Categories Page work again

VKormylo:
Fixed Categories page not found error. I refactored code so that it uses useQuery instead of useAxios for data fetching.
### Categories page:
image
### Subjects page:
image
### Subjects page filtered by category:
image

### DIFF
--- a/src/constants/request.ts
+++ b/src/constants/request.ts
@@ -52,6 +52,7 @@ export const URLs = {
   },
   subjects: {
     get: '/subjects',
+    getByCategoryId: '/categories/:categoryId/subjects',
     getNames: '/subjects/names'
   },
   cooperations: {
--- a/src/constants/request.ts
+++ b/src/constants/request.ts
@@ -52,7 +52,7 @@ export const URLs = {
   },
   subjects: {
     get: '/subjects',
-    getByCategoryId: '/categories/:categoryId/subjects',
+    getByCategoryId: '/categories/:id/subjects',
     getNames: '/subjects/names'
   },
   cooperations: {
--- a/src/hooks/use-load-more.tsx
+++ b/src/hooks/use-load-more.tsx
@@ -1,12 +1,18 @@
-import { useState, useMemo, useLayoutEffect, useCallback } from 'react'
+import {
+  useState,
+  useMemo,
+  useLayoutEffect,
+  useCallback,
+  useEffect
+} from 'react'
 
-import useAxios from '~/hooks/use-axios'
+import useQuery from '~/hooks/use-query'
 
 import { defaultResponses } from '~/constants'
-import { ServiceFunction, ItemsWithCount } from '~/types'
+import { ItemsWithCount, ServiceFunctionNew } from '~/types'
 
 interface UseLoadMoreProps<Data, Params> {
-  service: ServiceFunction<ItemsWithCount<Data>, Params>
+  service: ServiceFunctionNew<ItemsWithCount<Data>, Params>
   limit: number
   params?: Params
 }
@@ -36,20 +42,20 @@ const useLoadMore = <Data, Params>({
     setData([])
   }, [])
 
-  const { response, loading, fetchData } = useAxios<
-    ItemsWithCount<Data>,
-    Params
-  >({
-    service,
-    defaultResponse: defaultResponses.itemsWithCount,
-    fetchOnMount: false,
-    onResponse: handleResponse
+  const { data: response, isLoading: loading } = useQuery({
+    queryFn: () => service({ ...params, limit, skip } as Params),
+    queryKey: [`load-more`, params, limit, skip],
+    options: {
+      initialData: defaultResponses.itemsWithCount
+    }
   })
 
+  useEffect(() => {
+    handleResponse(response)
+  }, [response, handleResponse])
+
   useLayoutEffect(() => {
-    if (previousLimit === limit && !isFetched) {
-      void fetchData({ ...params, limit, skip } as Params)
-    } else {
+    if (previousLimit !== limit || isFetched) {
       resetData()
       setPreviousLimit(limit)
     }
@@ -58,7 +64,7 @@ const useLoadMore = <Data, Params>({
       // eslint-disable-next-line react-hooks/exhaustive-deps
       isFetched = true
     }
-  }, [fetchData, limit, previousLimit, resetData, skip, params])
+  }, [limit, previousLimit, resetData, skip, params])
 
   const isExpandable = useMemo(
     () => data.length < response.count && data.length > 0,
--- a/src/hooks/use-load-more.tsx
+++ b/src/hooks/use-load-more.tsx
@@ -44,7 +44,7 @@ const useLoadMore = <Data, Params>({
 
   const { data: response, isFetching: loading } = useQuery({
     queryFn: () => service({ ...params, limit, skip } as Params),
-    queryKey: [`load-more`, params, limit, skip],
+    queryKey: ['load-more', params, limit, skip],
     options: {
       initialData: defaultResponses.itemsWithCount
     }
--- a/src/hooks/use-load-more.tsx
+++ b/src/hooks/use-load-more.tsx
@@ -42,7 +42,7 @@ const useLoadMore = <Data, Params>({
     setData([])
   }, [])
 
-  const { data: response, isLoading: loading } = useQuery({
+  const { data: response, isFetching: loading } = useQuery({
     queryFn: () => service({ ...params, limit, skip } as Params),
     queryKey: [`load-more`, params, limit, skip],
     options: {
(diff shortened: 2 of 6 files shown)


SofiiaYevush (approved):

src/hooks/use-load-more.tsx
```diff
@@ -36,20 +42,20 @@ const useLoadMore = <Data, Params>({
...
-  const { response, loading, fetchData } = useAxios<
-    ItemsWithCount<Data>,
-    Params
-  >({
-    service,
-    defaultResponse: defaultResponses.itemsWithCount,
-    fetchOnMount: false,
-    onResponse: handleResponse
+  const { data: response, isFetching: loading } = useQuery({
+    queryFn: () => service({ ...params, limit, skip } as Params),
```
ShadowOfTheSpace, line 46: Why do you need to cast is to Params?

src/services/subject-service.ts
```diff
@@ -4,14 +4,30 @@ import { AxiosResponse } from 'axios'
...
+import { getFullUrl } from '~/utils/get-full-url'
 
 export const subjectService = {
-  getSubjects: (
-    params?: Pick<SubjectInterface, 'name'>,
-    categoryId?: string
-  ): Promise<AxiosResponse<ItemsWithCount<SubjectInterface>>> => {
-    const category = createUrlPath(URLs.categories.get, categoryId)
-    return axiosClient.get(`${category}${URLs.subjects.get}`, { params })
+  getSubjects: (params?: Pick<SubjectInterface, 'name' | 'category'>) => {
```
ShadowOfTheSpace: Why `params` is optional?
```suggestion
  getSubjects: (params: Pick<SubjectInterface, 'name' | 'category'>) => {
```

src/services/subject-service.ts
```diff
@@ -4,14 +4,30 @@ import { AxiosResponse } from 'axios'
...
+    const { category: { _id: categoryId } = {}, ...restParams } = params ?? {}
```
ShadowOfTheSpace:
```suggestion
    const { category, ...restParams } = params
```

src/constants/request.ts
```diff
@@ -52,6 +52,7 @@ export const URLs = {
   },
   subjects: {
     get: '/subjects',
+    getByCategoryId: '/categories/:categoryId/subjects',
```
ShadowOfTheSpace: You don't need to add named parameters when there is only one of them
```suggestion
    getByCategoryId: '/categories/:id/subjects',
```

src/hooks/use-load-more.tsx
```diff
@@ -36,20 +42,20 @@ const useLoadMore = <Data, Params>({
...
+    queryKey: [`load-more`, params, limit, skip],
```
ShadowOfTheSpace:
```suggestion
    queryKey: ['load-more', params, limit, skip],
```

src/pages/subjects/Subjects.tsx
```diff
@@ -78,9 +78,9 @@ const Subjects = () => {
   }
 
   const getSubjects = useCallback(
-    (data?: Pick<SubjectInterface, 'name'>) =>
-      subjectService.getSubjects(data, categoryId),
-    [categoryId]
+    (data?: Pick<SubjectInterface, 'name' | 'category'>) =>
```
ShadowOfTheSpace: Why is it optional here? You replaced useAxios with useQuery here, so optionality is redundant here
```suggestion
    (data: Pick<SubjectInterface, 'name' | 'category'>) =>
```

src/pages/subjects/Subjects.tsx
ShadowOfTheSpace, line 1: Please, check how does Subects API work, probably `name` is not needed for requests. Probably only categoryId is needed

src/services/subject-service.ts
```diff
@@ -4,14 +4,30 @@ import { AxiosResponse } from 'axios'
...
+      searchParameters: restParams
+    })
+
+    if (categoryId) {
+      resultUrl = getFullUrl({
+        pathname: URLs.subjects.getByCategoryId,
+        parameters: { categoryId },
+        searchParameters: restParams
+      })
+    }
```
ShadowOfTheSpace:
```suggestion
    if (category?._id) {
      resultUrl = getFullUrl({
        pathname: URLs.subjects.getByCategoryId,
        parameters: { id: category._id},
        searchParameters: restParams
      })
    }
```

src/types/services/types/services.types.ts
```diff
@@ -18,9 +18,12 @@ export type ServiceFunction<Response, Params = undefined> = (
...
 
-export type ServiceFunctionNew<Response, Params = undefined> = (
-  params: Params extends undefined ? undefined : Params
-) => Promise<Response>
+export type ServiceFunctionNew<
+  Response,
+  Params = undefined
+> = Params extends undefined
+  ? () => Promise<Response>
+  : (params: Params) => Promise<Response>
```
ShadowOfTheSpace: Please, make a rebase. This type is already as you need on develop branch

tests/unit/services/subject-service.spec.jsx
```diff
@@ -0,0 +1,81 @@
...
+    vi.clearAllMocks()
+  })
+
+  it('should find subjects by categoryId', async () => {
+    const getFullUrlSpy = vi.spyOn(getFullUrl, 'getFullUrl')
+
+    mockAxiosClient
+      .onGet(
+        new RegExp(
+          URLs.subjects.getByCategoryId.replace(':categoryId', mockCategoryId)
```
ShadowOfTheSpace: Don't forget to change `categoryId` to `id` here

ShadowOfTheSpace (changes requested):

src/hooks/use-load-more.tsx
VKormylo: Without explicit casting there is type errors. I also can change that to `service({ ...(params as Params), limit, skip })` as well. What do you think about that change?

src/pages/subjects/Subjects.tsx
VKormylo: `name` is also needed because we can filter by categoryId and search by subject name.

src/services/subject-service.ts
VKormylo: Yes, it shouldn't be optional, thank you. Fixed.

src/services/subject-service.ts
VKormylo: Fixed.

src/constants/request.ts
VKormylo: Fixed.

src/hooks/use-load-more.tsx
VKormylo: Fixed.

src/pages/subjects/Subjects.tsx
VKormylo: Yeah, true. Fixed.

src/services/subject-service.ts
VKormylo: Fixed.

src/types/services/types/services.types.ts
VKormylo: Ok, thanks. Done.

tests/unit/services/subject-service.spec.jsx
VKormylo: Yes, thanks. Fixed.

src/hooks/use-load-more.tsx
ShadowOfTheSpace: I got it, leave it as it its

src/pages/subjects/Subjects.tsx
```diff
@@ -78,9 +78,9 @@ const Subjects = () => {
   }
 
   const getSubjects = useCallback(
-    (data?: Pick<SubjectInterface, 'name'>) =>
-      subjectService.getSubjects(data, categoryId),
-    [categoryId]
+    (data: Pick<SubjectInterface, 'name' | 'category'>) =>
```
ShadowOfTheSpace: if only `categoryId` is needed, you can create type like 
`Pick<SubjectInterface, 'name'> & Record<'categoryId, string>`
You will have to pass just an `id` without mocking entire category.

ShadowOfTheSpace (changes requested):

src/pages/subjects/Subjects.tsx
VKormylo: I was thinking about alternative solution to that also, so thanks a lot. I will change that.

VKormylo: Fixed.

amoutens (approved):

ShadowOfTheSpace (approved):

PR #3160, opened 2025-01-29
Fixed profile editing

### LINKED ISSUE #3141: (SP: ) Fix update button on EditProfile page
Actual result: after providing changes to Edit Profile, switching to other Tab and click to Update button our changed data was updated, but on the other hand, after reloading the page info didn't update.

https://github.com/user-attachments/assets/bf83cc1a-d690-446a-bd59-8fcec711c115

PavloButynets:
This PR depends on [Backend PR #1180](https://github.com/ita-social-projects/SpaceToStudy-BackEnd/pull/1180).
Merge the backend PR first to avoid breaking changes.

When sending a request to the server, the entire `profileState` object was sent, even those fields that were not edited, so a validation error occurred on the server because some fields (for example, `professionalSummary`, `nativeLanguage`) were empty strings. I also found another bug, when a user wanted to clear `professionalSummary` or `nativeLanguage`, he could not do it, because a request was also sent with an empty string and an error was thrown on the server. I implemented the request in such a way that only those fields that have changed are sent, and if the field is an **empty string** (“”), then **null** is sent

- Before:

https://github.com/user-attachments/assets/e15c7bd9-4621-4490-b481-2c7078f81398

- After

https://github.com/user-attachments/assets/f010fb7c-0836-4c23-9d63-c7dc212b2914

### DIFF
--- a/src/pages/edit-profile/EditProfile.tsx
+++ b/src/pages/edit-profile/EditProfile.tsx
@@ -161,11 +161,16 @@ const EditProfile = () => {
     if (typeof photo === 'object' || photo === '') {
       dataToUpdate.photo = photo
     }
-
+    const dataWithoutEmptyStrings = Object.fromEntries(
+      Object.entries(dataToUpdate).map(([key, value]) => [
+        key,
+        value === '' ? null : value
+      ])
+    )
     await dispatch(
       updateUser({
         userId,
-        params: dataToUpdate
+        params: dataWithoutEmptyStrings
       })
     )
 
--- a/src/pages/edit-profile/EditProfile.tsx
+++ b/src/pages/edit-profile/EditProfile.tsx
@@ -36,6 +36,7 @@ import { authRoutes } from '~/router/constants/authRoutes'
 
 import { styles } from '~/pages/edit-profile/EditProfile.styles'
 import { getChangedFields } from '~/utils/get-changed-fields'
+import { replaceEmptyStringsWithNull } from '~/utils/replace-empty-strings-with-null'
 
 const EditProfile = () => {
   const [initialEditProfileState, setInitialEditProfileState] = useState<
@@ -95,10 +96,8 @@ const EditProfile = () => {
   }, [loading, profileState, initialEditProfileState])
 
   const changedFields = useMemo<Partial<EditProfileState>>(() => {
-    return getChangedFields(
-      initialEditProfileState,
-      profileState
-    ) as Partial<EditProfileState>
+    if (!profileState || !initialEditProfileState) return {}
+    return getChangedFields(initialEditProfileState, profileState)
   }, [profileState, initialEditProfileState])
   const isChanged = useMemo<boolean>(
     () => Object.keys(changedFields).length > 0,
@@ -161,12 +160,7 @@ const EditProfile = () => {
     if (typeof photo === 'object' || photo === '') {
       dataToUpdate.photo = photo
     }
-    const dataWithoutEmptyStrings = Object.fromEntries(
-      Object.entries(dataToUpdate).map(([key, value]) => [
-        key,
-        value === '' ? null : value
-      ])
-    )
+    const dataWithoutEmptyStrings = replaceEmptyStringsWithNull(dataToUpdate)
     await dispatch(
       updateUser({
         userId,
--- a/src/pages/edit-profile/EditProfile.tsx
+++ b/src/pages/edit-profile/EditProfile.tsx
@@ -96,7 +96,10 @@ const EditProfile = () => {
   }, [loading, profileState, initialEditProfileState])
 
   const changedFields = useMemo<Partial<EditProfileState>>(() => {
-    if (!profileState || !initialEditProfileState) return {}
+    if (!profileState || !initialEditProfileState) {
+      return {}
+    }
+
     return getChangedFields(initialEditProfileState, profileState)
   }, [profileState, initialEditProfileState])
   const isChanged = useMemo<boolean>(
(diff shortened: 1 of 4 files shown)


ArtemHolikov (approved):

SofiiaYevush (approved):

markgol777 (approved):

src/pages/edit-profile/EditProfile.tsx
```diff
@@ -102,36 +95,11 @@ const EditProfile = () => {
...
-      }
-
-      return changes
-    } else {
-      return {}
-    }
+    return getChangedFields(
+      initialEditProfileState,
+      profileState
+    ) as Partial<EditProfileState>
```
ShadowOfTheSpace: Why do you need type assertion here?

src/pages/edit-profile/EditProfile.tsx
```diff
@@ -193,13 +161,19 @@ const EditProfile = () => {
     if (typeof photo === 'object' || photo === '') {
       dataToUpdate.photo = photo
     }
-
+    const dataWithoutEmptyStrings = Object.fromEntries(
+      Object.entries(dataToUpdate).map(([key, value]) => [
+        key,
+        value === '' ? null : value
+      ])
+    )
```
ShadowOfTheSpace: You can extract that as a utility function, it will be used a lot in our project I think.

src/utils/get-changed-fields.ts
```diff
@@ -0,0 +1,26 @@
+type Nullable<T> = {
+  [K in keyof T]: T[K] | null
+}
```
ShadowOfTheSpace: You don't need that type here.

```suggestion
```

src/utils/get-changed-fields.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function getChangedFields<T>(
+  initialState: T | null,
+  currentState: T | null
+): Partial<Nullable<T>> {
```
ShadowOfTheSpace:
```suggestion
): Partial<T> {
```

src/utils/get-changed-fields.ts
```diff
@@ -0,0 +1,26 @@
+type Nullable<T> = {
+  [K in keyof T]: T[K] | null
+}
+
+export function getChangedFields<T>(
+  initialState: T | null,
+  currentState: T | null
```
ShadowOfTheSpace: I thought we discussed that before: you don't need to get changes of objects that have different types.
What is the difference between `{ field: 'value' }` and `null` and otherwise? I think it only matters to find the difference between equal types.

src/utils/get-changed-fields.ts
```diff
@@ -0,0 +1,26 @@
+type Nullable<T> = {
+  [K in keyof T]: T[K] | null
+}
+
+export function getChangedFields<T>(
```
ShadowOfTheSpace: Why you didn't specify a more detailed generic type here? Just `T` can be anything, I can pass here a plain string, array, or something else. I think we should make that forbidden here.

```suggestion
export function getChangedFields<T extends Record<string, unknown>>(
```

src/utils/get-changed-fields.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  const changes: Partial<Nullable<T>> = {}
+
+  Object.keys(currentState).forEach((key) => {
+    const typedKey = key as keyof T
+    const initialValue = initialState[typedKey]
+    const currentValue = currentState[typedKey]
+
+    if (
+      (currentValue !== undefined || initialValue !== undefined) &&
```
ShadowOfTheSpace: Why to we need to run `JSON.stringify` for undefined and defined fields?

ShadowOfTheSpace (changes requested):

src/utils/get-changed-fields.ts
PavloButynets: yes, I removed the null setting and forgot to remove this type

src/utils/get-changed-fields.ts
PavloButynets: sorry, now it isn't necessary

src/pages/edit-profile/EditProfile.tsx
```diff
@@ -102,36 +96,9 @@ const EditProfile = () => {
...
-      if (hasPhotoChanged) {
-        changes.photo = currentPhoto
-      }
-
-      return changes
-    } else {
-      return {}
-    }
+    if (!profileState || !initialEditProfileState) return {}
+    return getChangedFields(initialEditProfileState, profileState)
```
ShadowOfTheSpace:
```suggestion
    if (!profileState || !initialEditProfileState) { 
    return {}
    }
    
    return getChangedFields(initialEditProfileState, profileState)
```

src/pages/edit-profile/EditProfile.tsx
```diff
@@ -193,13 +160,14 @@ const EditProfile = () => {
     if (typeof photo === 'object' || photo === '') {
       dataToUpdate.photo = photo
     }
-
+    const dataWithoutEmptyStrings = replaceEmptyStringsWithNull(dataToUpdate)
```
ShadowOfTheSpace, line 167:
```suggestion

    const dataWithoutEmptyStrings = replaceEmptyStringsWithNull(dataToUpdate)
    
```

src/utils/get-changed-fields.ts
```diff
@@ -0,0 +1,19 @@
...
+  Object.keys(currentState).forEach((key) => {
+    const typedKey = key as keyof T
+    const initialValue = initialState[typedKey]
+    const currentValue = currentState[typedKey]
+
+    if (JSON.stringify(initialValue) !== JSON.stringify(currentValue)) {
+      changes[typedKey] = currentValue
+    }
+  })
+  return changes
```
ShadowOfTheSpace, line 17:
```suggestion

  return changes
```

ShadowOfTheSpace (approved):
Approved, but fix those comments about code style, please)

PR #3186, opened 2025-02-07
Replace useAxios with useQuery in the AttachmentsContainer

### LINKED ISSUE #3110: Replace useAxios with useQuery in the AttachmentsContainer
Replace the useAxios hook with a custom useQuery (useMutation) hook on the CooperationActivities page to improve data fetching, caching, and error handling

- Identify where useAxios is used
- Replaces all of the instances of useAxios with useQuery
- Update tests if necessary
- Ensure that everything works as intended
- Make sure that isLoading or isFetching is used correctly

luiqor:
### Replace useAxios with useQuery in the AttachmentsContainer

- Replace useAxios with useQuery for loading attachements
- Replace useAxios with useMutation for updating attachements

https://github.com/user-attachments/assets/c1199cbf-331a-4f54-8aac-a556664b3087

- Replace useAxios with useMutation for creating attachements

https://github.com/user-attachments/assets/a33ae3a3-4353-4b87-8e09-b9e86d37c4f2

- Renamed methods in ResourceService for attachments as there are only methods using baseQuery used around app
- Replaced onCreateAttachmentsError, onResponseError with usage of handleErrorAlert (custom useSnackbarAlert hook) 

- Add tests for resource-service
- Update tests for AddAttachmentCategoryModal

### DIFF
--- a/src/containers/my-resources/add-attachment-category-modal/AddAttachmentCategoryModal.tsx
+++ b/src/containers/my-resources/add-attachment-category-modal/AddAttachmentCategoryModal.tsx
@@ -22,13 +22,13 @@ import {
 interface AddAttachmentCategoryModalProps {
   closeModal: () => void
   attachment: Attachment
-  updateAttachmentCategory: (params: UpdateAttachmentParams) => void
+  onAttachmentUpdate: (params: UpdateAttachmentParams) => void
 }
 
 const AddAttachmentCategoryModal: FC<AddAttachmentCategoryModalProps> = ({
   closeModal,
   attachment,
-  updateAttachmentCategory
+  onAttachmentUpdate
 }) => {
   const { t } = useTranslation()
 
@@ -39,7 +39,10 @@ const AddAttachmentCategoryModal: FC<AddAttachmentCategoryModalProps> = ({
       initialValues: getInitialValues(attachment),
       onSubmit: () => {
         setLoading(true)
-        updateAttachmentCategory(data)
+        onAttachmentUpdate({
+          id: attachment._id,
+          category: data.category
+        })
         setLoading(false)
         closeModal()
       }
@@ -84,7 +87,7 @@ const AddAttachmentCategoryModal: FC<AddAttachmentCategoryModalProps> = ({
           {t('common.cancel')}
         </Button>
         <Button
-          disabled={!!errors.fileName}
+          disabled={Boolean(errors.fileName) || !data.category}
           loading={loading}
           sx={styles.saveBtn}
           type={ButtonTypeEnum.Submit}
--- a/src/containers/my-resources/add-attachment-category-modal/AddAttachmentCategoryModal.tsx
+++ b/src/containers/my-resources/add-attachment-category-modal/AddAttachmentCategoryModal.tsx
@@ -1,4 +1,3 @@
-import { FC, SyntheticEvent, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import Box from '@mui/material/Box'
 import Typography from '@mui/material/Typography'
@@ -25,31 +24,27 @@ interface AddAttachmentCategoryModalProps {
   onAttachmentUpdate: (params: UpdateAttachmentParams) => void
 }
 
-const AddAttachmentCategoryModal: FC<AddAttachmentCategoryModalProps> = ({
+const AddAttachmentCategoryModal: React.FC<AddAttachmentCategoryModalProps> = ({
   closeModal,
   attachment,
   onAttachmentUpdate
 }) => {
   const { t } = useTranslation()
 
-  const [loading, setLoading] = useState<boolean>(false)
-
   const { data, errors, handleNonInputValueChange, handleBlur, handleSubmit } =
     useForm<UpdateAttachmentParams>({
       initialValues: getInitialValues(attachment),
       onSubmit: () => {
-        setLoading(true)
         onAttachmentUpdate({
           id: attachment._id,
           category: data.category
         })
-        setLoading(false)
         closeModal()
       }
     })
 
   const onCategoryChange = (
-    _: SyntheticEvent,
+    _: React.SyntheticEvent,
     value: CategoryNameInterface | null
   ) => {
     handleNonInputValueChange('category', value?._id ?? null)
@@ -88,7 +83,6 @@ const AddAttachmentCategoryModal: FC<AddAttachmentCategoryModalProps> = ({
         </Button>
         <Button
           disabled={Boolean(errors.fileName) || !data.category}
-          loading={loading}
           sx={styles.saveBtn}
           type={ButtonTypeEnum.Submit}
         >
--- a/src/containers/my-resources/add-attachment-category-modal/AddAttachmentCategoryModal.tsx
+++ b/src/containers/my-resources/add-attachment-category-modal/AddAttachmentCategoryModal.tsx
@@ -22,9 +22,7 @@ import {
 interface AddAttachmentCategoryModalProps {
   closeModal: () => void
   attachment: Attachment
-  updateAttachmentCategory: (
-    params?: UpdateAttachmentParams | undefined
-  ) => Promise<void>
+  updateAttachmentCategory: (params: UpdateAttachmentParams) => void
 }
 
 const AddAttachmentCategoryModal: FC<AddAttachmentCategoryModalProps> = ({
@@ -39,9 +37,9 @@ const AddAttachmentCategoryModal: FC<AddAttachmentCategoryModalProps> = ({
   const { data, errors, handleNonInputValueChange, handleBlur, handleSubmit } =
     useForm<UpdateAttachmentParams>({
       initialValues: getInitialValues(attachment),
-      onSubmit: async () => {
+      onSubmit: () => {
         setLoading(true)
-        await updateAttachmentCategory(data)
+        updateAttachmentCategory(data)
         setLoading(false)
         closeModal()
       }
--- a/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
+++ b/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
@@ -60,7 +60,6 @@ const AttachmentsContainer = () => {
     [itemsPerPage, page, sort, searchFileName, selectedItems]
   )
 
-  // useMutation hook
   const deleteAttachment = useCallback(
     (id?: string) => ResourceService.deleteAttachment(id ?? ''),
     []
@@ -149,7 +148,7 @@ const AttachmentsContainer = () => {
   const props = {
     columns: columnsToShow,
     data: {
-      response: loadedAttachments ?? defaultResponses.itemsWithCount,
+      response: loadedAttachments,
       getData: handleRefetchAttachments
     },
     services: { deleteService: deleteAttachment },
--- a/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
+++ b/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
@@ -98,11 +98,7 @@ const AttachmentsContainer = () => {
   })
 
   const onEdit = (id: string) => {
-    if (!loadedAttachments) {
-      return
-    }
-
-    const attachment = loadedAttachments.items.find((item) => item._id === id)
+    const attachment = loadedAttachments!.items.find((item) => item._id === id)
 
     const handleConfirm = () =>
       openModal({
@@ -127,11 +123,7 @@ const AttachmentsContainer = () => {
   }
 
   const onAddCategory = (id: string) => {
-    if (!loadedAttachments) {
-      return
-    }
-
-    const attachment = loadedAttachments.items.find((item) => item._id === id)
+    const attachment = loadedAttachments!.items.find((item) => item._id === id)
 
     openModal({
       component: (
--- a/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
+++ b/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
@@ -11,7 +11,6 @@ import MyResourcesTable from '~/containers/my-resources/my-resources-table/MyRes
 import Loader from '~/components/loader/Loader'
 import useSort from '~/hooks/table/use-sort'
 import useBreakpoints from '~/hooks/use-breakpoints'
-import useAxios from '~/hooks/use-axios'
 import usePagination from '~/hooks/table/use-pagination'
 import AddDocuments from '~/containers/add-documents/AddDocuments'
 
@@ -23,12 +22,11 @@ import {
   removeColumnRules
 } from '~/containers/my-resources/attachments-container/AttachmentsContainer.constants'
 import {
-  ItemsWithCount,
-  Attachment,
-  ErrorResponse,
-  ResourcesTabsEnum,
-  ButtonVariantEnum,
-  CooperationSliceAttachment
+  type ItemsWithCount,
+  type Attachment,
+  type ErrorResponse,
+  type CooperationSliceAttachment,
+  ResourcesTabsEnum
 } from '~/types'
 import { adjustColumns, getScreenBasedLimit } from '~/utils/helper-functions'
 import { styles } from '~/containers/my-resources/attachments-container/AttachmentsContainer.styles'
@@ -105,11 +103,6 @@ const AttachmentsContainer = () => {
     queryKey: ['attachments']
   })
 
-  const createAttachments = useCallback(
-    (data?: FormData) => ResourceService.createAttachments(data),
-    []
-  )
-
   const onCreateAttachmentsError = (error?: ErrorResponse) => {
     dispatch(
       openAlert({
@@ -119,18 +112,11 @@ const AttachmentsContainer = () => {
     )
   }
 
-  // TODO: useMutation
-  const { fetchData: fetchCreateAttachment } = useAxios({
-    service: createAttachments,
-    fetchOnMount: false,
-    defaultResponse: null,
-    onResponseError: onCreateAttachmentsError
+  const { mutate: mutateCreateAttachment } = useMutation({
+    mutationFn: ResourceService.createAttachments,
+    onError: onCreateAttachmentsError,
+    queryKey: ['attachments']
   })
-
-  const uploadFile = async (data: FormData) => {
-    await fetchCreateAttachment(data)
-    await fetchAttachments()
-  }
 
   const onEdit = (id: string) => {
     const attachment = response.items.find((item) => item._id === id)
@@ -197,12 +183,11 @@ const AttachmentsContainer = () => {
       button={
         <AddDocuments
           buttonText={t('myResourcesPage.attachments.addBtn')}
-          fetchData={uploadFile}
+          fetchData={mutateCreateAttachment}
           formData={formData}
           icon={<AddIcon sx={styles.addAttachmentIcon} />}
           removePreviousFiles
           sx={styles.addAttachmentBtn}
-          variant={ButtonVariantEnum.Contained}
         />
       }
       fetchData={fetchAttachments}
--- a/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
+++ b/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
@@ -75,7 +75,13 @@ const AttachmentsContainer = () => {
     refetch: refetchAttachments,
     error: fetchAttachmentsError
   } = useQuery<ItemsWithCount<Attachment>>({
-    queryKey: ['attachments'],
+    queryKey: [
+      'attachments',
+      page,
+      sort,
+      searchFileName.current,
+      selectedItems
+    ],
     queryFn: getAttachments,
     options: {
       initialData: defaultResponses.itemsWithCount
--- a/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
+++ b/src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
@@ -65,7 +65,7 @@ const AttachmentsContainer = () => {
 
   const getAttachments = useCallback(
     () =>
-      ResourceService.getAttachmentsQuery({
+      ResourceService.getAttachments({
         limit: itemsPerPage,
         skip: (page - 1) * itemsPerPage,
         sort,
@@ -97,8 +97,8 @@ const AttachmentsContainer = () => {
     await refetchAttachments()
   }
 
-  const { mutate: mutateAttachment } = useMutation({
-    mutationFn: ResourceService.updateAttachmentQuery,
+  const { mutate: updateAttachment } = useMutation({
+    mutationFn: ResourceService.updateAttachment,
     onError: onResponseError,
     queryKey: ['attachments']
   })
@@ -112,8 +112,8 @@ const AttachmentsContainer = () => {
     )
   }
 
-  const { mutate: mutateCreateAttachment } = useMutation({
-    mutationFn: ResourceService.createAttachments,
+  const { mutate: createAttachment } = useMutation({
+    mutationFn: ResourceService.createAttachment,
     onError: onCreateAttachmentsError,
     queryKey: ['attachments']
   })
@@ -127,7 +127,7 @@ const AttachmentsContainer = () => {
           <EditAttachmentModal
             attachment={attachment as CooperationSliceAttachment}
             closeModal={closeModal}
-            onAttachmentUpdate={mutateAttachment}
+            onAttachmentUpdate={updateAttachment}
           />
         )
       })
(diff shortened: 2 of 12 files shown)


src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
```diff
@@ -82,72 +65,51 @@ const AttachmentsContainer = () => {
...
-    response,
-    loading,
-    fetchData: fetchAttachments
-  } = useAxios<ItemsWithCount<Attachment>, GetResourcesParams>({
-    service: getAttachments,
-    defaultResponse: defaultResponses.itemsWithCount,
-    onResponseError
+    data: loadedAttachments,
+    isLoading: isLoadingAttachments,
+    refetch: refetchAttachments,
```
ShadowOfTheSpace: Can we get rid from refetch with keys invalidating?

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
```diff
@@ -82,72 +65,51 @@ const AttachmentsContainer = () => {
...
+      'attachments',
+      page,
+      sort,
+      searchFileName.current,
+      selectedItems
+    ],
+    queryFn: getAttachments,
+    options: {
+      initialData: defaultResponses.itemsWithCount
+    }
```
ShadowOfTheSpace: Can we show loader instead of defaultValues? I think it's a better UX for user to show loader instead of irrelevant data, and also this approach prevents layout shifting

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
```diff
@@ -82,72 +65,51 @@ const AttachmentsContainer = () => {
...
   })
 
-  const onAttachmentUpdate = useCallback(
-    () => void fetchAttachments(),
-    [fetchAttachments]
-  )
+  const handleRefetchAttachments = async (): Promise<void> => {
+    await refetchAttachments()
+  }
```
ShadowOfTheSpace: Can you pass just a reference to `refetchAttachments` instead of wrapper?

src/services/resource-service.ts
```diff
@@ -122,11 +122,7 @@ export const ResourceService = {
       })
     })
   },
-  getAttachments: async (
-    params?: GetResourcesParams
-  ): Promise<AxiosResponse<ItemsWithCount<Attachment>>> =>
-    await axiosClient.get(URLs.resources.attachments.get, { params }),
-  getAttachmentsQuery: (params?: GetResourcesParams) => {
+  getAttachments: (params?: GetResourcesParams) => {
```
ShadowOfTheSpace: Should those `params` be required?
```suggestion
  getAttachments: (params: GetResourcesParams) => {
```

src/services/resource-service.ts
```diff
@@ -158,9 +148,11 @@ export const ResourceService = {
...
     )
   },
-  createAttachments: (data?: FormData): Promise<AxiosResponse> => {
-    return axiosClient.post(URLs.attachments.post, data, {
-      headers: { 'Content-Type': 'multipart/form-data' }
+  createAttachment: (data: FormData) => {
+    return baseService.request<Attachment>({
+      method: 'POST',
+      url: URLs.resources.attachments.post,
+      data: data
```
ShadowOfTheSpace:
```suggestion
      data,
```

src/services/resource-service.ts
```diff
@@ -158,9 +148,11 @@ export const ResourceService = {
       createUrlPath(URLs.resources.attachments.delete, id)
     )
   },
-  createAttachments: (data?: FormData): Promise<AxiosResponse> => {
-    return axiosClient.post(URLs.attachments.post, data, {
-      headers: { 'Content-Type': 'multipart/form-data' }
+  createAttachment: (data: FormData) => {
+    return baseService.request<Attachment>({
+      method: 'POST',
```
ShadowOfTheSpace, line 173: Does it work without `headers`?

tests/unit/services/resource-service.spec.jsx
```diff
@@ -93,32 +93,77 @@ describe('resourseService tests', () => {
...
+      link: '1722535882408-test.pdf',
+      size: 15069,
+      resourceType: 'Attachment'
+    }
+
+    mockAxiosClient.onPatch().reply((config) => {
+      expect(config.url).toBe(`/attachments/${attachmentId}`)
+
+      return [200, mockAttachmentResponse]
+    })
```
ShadowOfTheSpace: Message me, if this won't work
```suggestion
    mockAxiosClient.onPatch(`/attachments/${attachmentId}`).reply(200, mockAttachmentResponse)
```

tests/unit/services/resource-service.spec.jsx
ShadowOfTheSpace, line 1: Please, import `URL` constant here and use it in every axiosMock. If we change some url - your test will still work, but component itself won't.

ShadowOfTheSpace (changes requested):

src/services/resource-service.ts
luiqor: Yes, absolutely in the same way as with them. That is interesting. I just simply forgot about headers when was working on rewriting that request. What do you think, should I add `headers` back?

src/services/resource-service.ts
luiqor: Good thing you noticed that. Tested on backend. It appeared that fileName param is actually required. But it was optional in frontend code (GetResourcesParams type). So now params are obviously required here

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: Than that leads to problems with types:
```
Type '(options?: RefetchOptions | undefined) => Promise<QueryObserverResult<ItemsWithCount<Attachment>, ResponseError>>' is not assignable to type '() => Promise<void>'.
  Type 'Promise<QueryObserverResult<ItemsWithCount<Attachment>, ResponseError>>' is not assignable to type 'Promise<void>'.
    Type 'QueryObserverResult<ItemsWithCount<Attachment>, ResponseError>' is not assignable to type 'void'.
      Type 'QueryObserverRefetchErrorResult<ItemsWithCount<Attachment>, ResponseError>' is not assignable to type 'void'.ts(2322)
AddResourceWithInput.tsx(31, 3): The expected type comes from property 'fetchData' which is declared here on type 'IntrinsicAttributes & AddResourceWithInputProps'
(property) AddResourceWithInputProps.fetchData: () => Promise<void>
```

AddResourceWithInput and MyResourcesTable expect `(params?: GetResourcesParams) => Promise<void> | void` and `() => Promise<void>`. Both components used in many places across our app

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: We [already have a loader](https://github.com/ita-social-projects/SpaceToStudy-Client/pull/3186/files/697913b3e9f34c5878b7ae83db83a842c1286ae6#diff-4dc98fdee37d3fed52b9492a3d6c0b7acde3f4651cf987403002e658e4d2aa0dL224) that is shown when attachments are loading (`isLoadingAttachments`):
image

I think that loader should be shown only to show that loading is in the process. But for other cases, like errors, there should be just initialValues and message (error snackbar in that case):

image

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: I tried to use invalidateQueries like that:
```
// AttachmentsContainer.tsx

import { queryClient } from '~/plugins/queryClient'
// ...
  const invalidateAttachments = async () => {
    await queryClient.invalidateQueries({
      queryKey: [
        'attachments',
        page,
        sort,
        searchFileName.current,
        selectedItems
      ]
    })
  }
  ```
and than pass invalidateAttachments to `MyResourcesTable` and `AddResourceWithInput`, but it doesn't work. I always see `initialValues` then:
image

Or did you mean something else?

Comment (ShadowOfTheSpace) [reply src/containers/my-resources/attachments-container/AttachmentsContainer.tsx]: If we have `initialData` inside `useQuery` and some `staleTime` - initial data will be used as a cached value, and since it's already in cashe, `useQuery` won't run request to fetch the data.
So if there is an `initailData` inside `useQuery` - invalidation won't work corectly :(

File: src/containers/my-resources/attachments-container/AttachmentsContainer.tsx (hunk 2 above)
Comment (ShadowOfTheSpace) [reply src/containers/my-resources/attachments-container/AttachmentsContainer.tsx]: It's very nice that you thought about error state, really. It's a good idea to show `initialData` if there is an error, but if you use `initialData` - cashing and invalidating behaviour won't work corectly.
So I think we need to add default values only if there is an error, otherwise - show loader without initial values and layout shifts.

File: src/containers/my-resources/attachments-container/AttachmentsContainer.tsx (hunk 3 above)
Comment (ShadowOfTheSpace) [reply src/containers/my-resources/attachments-container/AttachmentsContainer.tsx]: So refetch function have some parameters and they're conflicting with `params` of callback inside props?

File: src/services/resource-service.ts (hunk 6 above)
Comment (ShadowOfTheSpace) [reply src/services/resource-service.ts:173]: If headers is not required for that request - we can easily skip them and don't add them back

File: src/containers/my-resources/attachments-container/AttachmentsContainer.tsx [hunk 8]
```diff
@@ -82,72 +65,51 @@ const AttachmentsContainer = () => {
...
-    fetchData: fetchAttachments
-  } = useAxios<ItemsWithCount<Attachment>, GetResourcesParams>({
-    service: getAttachments,
-    defaultResponse: defaultResponses.itemsWithCount,
-    onResponseError
+    data: loadedAttachments,
+    isLoading: isLoadingAttachments,
+    refetch: refetchAttachments,
+    error: attachmentsLoadError
+  } = useQuery<ItemsWithCount<Attachment>>({
```
ShadowOfTheSpace: You can remove those types since we don't have requirements for mandatory typing in our project. `useQuery` will be typed by `baseService` and `queryFn` itself
```suggestion
  } = useQuery({
```

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
ShadowOfTheSpace: Try to get rid of `initialData` and use them only if there is an error. Instead add `staleTime: Infinity` option to configure caching behavior

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
```diff
@@ -217,10 +181,16 @@ const AttachmentsContainer = () => {
     />
   )
 
+  useEffect(() => {
+    if (attachmentsLoadError) {
+      handleErrorAlert(attachmentsLoadError as ErrorResponse)
```
ShadowOfTheSpace: I think that `ErrorResponse` (type) and `ResponseError` (exception that is thrown inside `baseService` are incompatible.
Check, please, `handleErrorAlert`, can you add `ResponseError` option to it?

tests/unit/containers/my-resources/AddAttachmentCategoryModal.spec.jsx
```diff
@@ -25,6 +25,15 @@ const attachmentMock = {
...
 
+const changeCategory = (autocomplete, categoryName) => {
+  fireEvent.click(autocomplete)
+  fireEvent.change(autocomplete, {
+    target: { value: categoryName }
+  })
+  fireEvent.keyDown(autocomplete, { key: 'ArrowDown' })
+  fireEvent.keyDown(autocomplete, { key: 'Enter' })
+}
+
```
ShadowOfTheSpace: We have `selectOption` inside `test-utils`. Does it work for this test?

tests/unit/containers/my-resources/AddAttachmentCategoryModal.spec.jsx
```diff
@@ -52,26 +61,32 @@ describe('AddAttachmentCategoryModal component', () => {
     expect(title).toBeInTheDocument()
   })
 
-  it('should render save button and click on it', () => {
+  it('should render save button and click on it if category was changed', async () => {
```
ShadowOfTheSpace, line 57: Not async currently
```suggestion
  it('should render save button and click on it if category was changed', () => {
```

tests/unit/services/resource-service.spec.jsx
```diff
@@ -156,32 +156,69 @@ describe('resourseService tests', () => {
...
+    const updatedAttachmentResponse = await ResourceService.updateAttachment({
+      ...attachment,
+      id: attachmentId
+    })
+
+    expect(updatedAttachmentResponse).toEqual(mockAttachmentResponse)
+  })
+
+  it('should get attachements' , async () => {
+
```
ShadowOfTheSpace:
```suggestion
```

ShadowOfTheSpace (changes requested):

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: Sorry, my inattention here. ResponseError option already was for handleErrorAlert. Removed assertion

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: Params can be undefined. So I think the problem is in what refetchAttachments returns
```
const refetchAttachments: (options?: RefetchOptions) => Promise<QueryObserverResult<ItemsWithCount<Attachment>, ResponseError>>
```

luiqor: I've just replaced refecthing logic with invalidateQueries and removed `handleRefetchAttachments`

luiqor:
New previews (after replacing refetch with invalidatequeries):

https://github.com/user-attachments/assets/195450e5-e3de-4cfe-ae63-3b3343882efe


https://github.com/user-attachments/assets/901d5b3c-1509-4827-b982-2d8488fe859d


https://github.com/user-attachments/assets/a32ed19c-862b-4e53-84bf-5579ccc38097

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: Thank you for your comments regarding caching! Did that and made some new previews: [comment](https://github.com/ita-social-projects/SpaceToStudy-Client/pull/3186#issuecomment-2658595911). Please, write me back if you meant something different about loader `without layout shifts`

tests/unit/containers/my-resources/AddAttachmentCategoryModal.spec.jsx
luiqor: Added selectOption (it is async), so returned async for test back

src/containers/my-resources/add-attachment-category-modal/AddAttachmentCategoryModal.tsx
```diff
@@ -39,9 +37,12 @@ const AddAttachmentCategoryModal: FC<AddAttachmentCategoryModalProps> = ({
...
       initialValues: getInitialValues(attachment),
-      onSubmit: async () => {
+      onSubmit: () => {
         setLoading(true)
-        await updateAttachmentCategory(data)
+        onAttachmentUpdate({
+          id: attachment._id,
+          category: data.category
+        })
         setLoading(false)
```
ShadowOfTheSpace: `onAttachmentUpdate` is a synchronous function, so you won't see loader probably.

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
```diff
@@ -164,14 +123,14 @@ const AttachmentsContainer = () => {
   }
 
   const onAddCategory = (id: string) => {
-    const attachment = response.items.find((item) => item._id === id)
+    const attachment = loadedAttachments!.items.find((item) => item._id === id)
```
ShadowOfTheSpace, line 120: Are you sure that `!` is safe here?

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
```diff
@@ -82,72 +61,52 @@ const AttachmentsContainer = () => {
...
   })
 
-  const uploadFile = async (data: FormData) => {
-    await fetchCreateAttachment(data)
-    await fetchAttachments()
-  }
-
   const onEdit = (id: string) => {
-    const attachment = response.items.find((item) => item._id === id)
+    const attachment = loadedAttachments!.items.find((item) => item._id === id)
```
ShadowOfTheSpace, line 95: Are you sure that ! is safe here?

tests/unit/containers/AttachmentsContainer/AttachmentsContainer.spec.jsx
```diff
@@ -54,19 +54,19 @@ const attachmentMockData = {
   items: responseItemsMock
 }
 
+vi.mock('~/services/resource-service', () => ({
+  ResourceService: {
+    getAttachments: vi.fn(() => Promise.resolve(attachmentMockData))
+  }
+}))
+
```
ShadowOfTheSpace: Try to mock just `axiosClient`. I know that mocking something else can bring a lot of problems in future.

tests/unit/containers/my-resources/AddAttachmentCategoryModal.spec.jsx
```diff
@@ -52,26 +56,35 @@ describe('AddAttachmentCategoryModal component', () => {
...
-    fireEvent.change(categoryDropbox, {
-      target: { value: categoriesNamesMock[1].name }
-    })
-    fireEvent.keyDown(categoryDropbox, { key: 'ArrowDown' })
-    fireEvent.keyDown(categoryDropbox, { key: 'Enter' })
+    expect(saveBtn).toBeInTheDocument()
+
+    fireEvent.click(saveBtn)
+
+    expect(updateAttachmentCategory).not.toHaveBeenCalled()
```
ShadowOfTheSpace, line 79: Try to check if the button is disabled instead of checking if the callback is called. Removing pointer events from a button when it is in a disabled state is a button functionality. Your functionality and logic are to make this button disabled, so you better check that instead.
But also you can check both, so if something happens with the button - you will see it in your test. And also the same about your logic - if something breaks your logic with the button disabling - you also will see it in this test.

tests/unit/containers/my-resources/AddAttachmentCategoryModal.spec.jsx
```diff
@@ -52,26 +56,35 @@ describe('AddAttachmentCategoryModal component', () => {
...
+  })
+
+  it('should change category', async () => {
+    const categoryDropbox = screen.getByRole('combobox')
 
+    await selectOption(categoryDropbox, categoriesNamesMock[1].name)
     expect(categoryDropbox.value).toBe(categoriesNamesMock[1].name)
```
ShadowOfTheSpace: Expecting `option.value` is already inside `selectOption`, so you can remove that

tests/unit/services/resource-service.spec.jsx
```diff
@@ -156,32 +156,68 @@ describe('resourseService tests', () => {
...
+    const mockAttachmentResponse = {
+      ...attachment,
+      _id: attachmentId,
+      link: '1722535882408-test.pdf',
+      size: 15069,
+      resourceType: 'Attachment'
+    }
+
+    mockAxiosClient
+      .onPatch(RegExp(URLs.resources.attachments.patch.replace(':id', attachmentId)))
```
ShadowOfTheSpace: You should add `RegExp` only when you have `searchParameters` in the URL because mocked `axiosClient` expects a URL with `searchParameters`, so you should add regex just to not mock those parameters. Otherwise - just remove `RegExp` and check if the test works without it.
```suggestion
      .onPatch(URLs.resources.attachments.patch.replace(':id', attachmentId))
```

tests/unit/services/resource-service.spec.jsx
```diff
@@ -156,32 +156,68 @@ describe('resourseService tests', () => {
...
+      {
+        _id: '6255bc080a75adf9223df444',
+        link: '1722535882408-test.jpg',
+        resourceType: 'Attachment',
+        description: 'Attachment description',
+        category: '8655bc080a75adf9223df444'
+      }
+    ]
+
+    mockAxiosClient.onGet(URLs.resources.attachments.get).reply(200, mockResponse)
```
ShadowOfTheSpace, line 235: Don't we have some default `searchParameters` there (like sorting by created at field, or something like that)?

ShadowOfTheSpace (changes requested):

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
ShadowOfTheSpace: Thanks, you did a great job with caching in this task! ❤️

About `layout shifts`: that thing is when your component "blinks" very fast and shows different things instead of a loader. For example, initial values were previously shown here after a short time of showing loader (loader, initial values with the message that `You have no attachments yet` and after that immediately correct (real) values). I think it's an annoying behavior for users because a lot of content `shifts`.
Also, there are more `layout shifts` when we are talking about images (the user visits some page, there is a big picture, but it's not loaded yet, and if no proper aspect ratio is set, all content that was under the image - will be shown on the image place, and after image will be loaded - all that content will fastly be moved down and the image will take full place).

I think there is a good [video](https://www.youtube.com/watch?v=Z6WiGWDU0nU&ab_channel=GoogleSearchCentral) about it, check this out!

ShadowOfTheSpace:
> New previews (after replacing refetch with invalidatequeries):
> 
>  artifictial-timeout.mp4 
>  on-error.mp4 
>  functionality-check.mp4

Great examples, but if you can, try to add a `network` tab in the video. It would be very useful to see `caching` in action. Also, try to go to different pages to fully demonstrate caching features.
Thanks again!

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: We only use this function in table in the row of particular attachment. If attachment won't be loaded - than it would be impossible to trigger this action

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: [same reason](https://github.com/ita-social-projects/SpaceToStudy-Client/pull/3186#discussion_r1958514474)

tests/unit/containers/my-resources/AddAttachmentCategoryModal.spec.jsx
luiqor: I removed this test, as it doesn't make sense. Seems like it`s author initial purpuse was just to check if dropbox works

tests/unit/services/resource-service.spec.jsx
luiqor: Where exactly? We don't have them straight in resource-service, but we have sort of "default" search params defined inside AttachmentsContainer (getAttachments fucntion)

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
luiqor: Got it. Tkank you!

luiqor:
> Great examples, but if you can, try to add a `network` tab in the video. It would be very useful to see `caching` in action. Also, try to go to different pages to fully demonstrate caching features. Thanks again!

💠 New demo of caching

slow 4g network throttling is enabled in first video

https://github.com/user-attachments/assets/ff37edbf-affe-496f-8bb8-35333668e225

https://github.com/user-attachments/assets/273b7d21-7c7c-4b66-97e4-f1e50788bab6

tests/unit/services/resource-service.spec.jsx
ShadowOfTheSpace: I got it, you are right!

In this test, there are no search params, but we have them inside other components that are using this service, so it's impossible to test something there until you mock axiosClient with RegExp URL (just to ignore those search params)

src/containers/my-resources/attachments-container/AttachmentsContainer.tsx
```diff
@@ -200,15 +162,14 @@ const AttachmentsContainer = () => {
...
           icon={<AddIcon sx={styles.addAttachmentIcon} />}
+          onCreateDocument={handleCreateAttachment}
           removePreviousFiles
           sx={styles.addAttachmentBtn}
-          variant={ButtonVariantEnum.Contained}
         />
       }
-      fetchData={fetchAttachments}
+      fetchData={invalidateAttachments}
       placeholder={'myResourcesPage.attachments.searchInput'}
```
ShadowOfTheSpace:
```suggestion
      placeholder='myResourcesPage.attachments.searchInput'
```

ShadowOfTheSpace (approved):
Great job, thanks!

Mike-Popovych (approved):

markgol777 (approved):

Issue #3181, opened 2025-02-07
Replace useAxios with useQuery in the MyResourcesTable and all related services passed as props


luiqor:
Replace the useAxios hook with a custom useQuery (useMutation) hook in current component to improve data fetching, caching, and error handling

- Identify where useAxios is used
- Replaces all of the instances of useAxios with useQuery
- Update tests if necessary
- Ensure that everything works as intended
- Make sure that isLoading or isFetching is used correctly

Related delete services passed as props:
![Image](https://github.com/user-attachments/assets/f4d14eb1-16ce-4eeb-95cf-304cc4af1ca8)

PR #3192, opened 2025-02-11
Replaced useAxios with useQuery in LessonsContainer

### LINKED ISSUE #3182: (SP: 3) Replace useAxios with useQuery in the LessonsContainer
Replace the useAxios hook with a custom useQuery (useMutation) hook in current component to improve data fetching, caching, and error handling

- Identify where useAxios is used
- Replaces all of the instances of useAxios with useQuery
- Update tests if necessary
- Ensure that everything works as intended
- Make sure that isLoading or isFetching is used correctly

Pay attention that delete passed as prop to MyResourcesTable will be handled in #3181

SofiiaYevush:
- The custom hook `useQuery` from the use-query library replaces `useAxios,` handling data fetching, loading, and error states more declaratively.
- The response object is now destructured with `useQuery’s` return values (`data, isLoading, error`), simplifying the data management.
- Updated and added new tests.

image

**Issue: #3182**

### DIFF
--- a/src/containers/my-resources/lessons-container/LessonsContainer.tsx
+++ b/src/containers/my-resources/lessons-container/LessonsContainer.tsx
@@ -85,7 +85,7 @@ const LessonsContainer = () => {
     if (!lessons) {
       return
     }
-    const resource = lessons.items.find((item) => item._id === id)
+    const resource = lessons.items?.find((item) => item._id === id)
     openModal({
       component: (
         <ChangeResourceConfirmModal
--- a/src/containers/my-resources/lessons-container/LessonsContainer.tsx
+++ b/src/containers/my-resources/lessons-container/LessonsContainer.tsx
@@ -10,8 +10,9 @@ import useSort from '~/hooks/table/use-sort'
 import useBreakpoints from '~/hooks/use-breakpoints'
 import useQuery from '~/hooks/use-query'
 import usePagination from '~/hooks/table/use-pagination'
+import useSnackbarAlert from '~/hooks/use-snackbar-alert'
 
-import { defaultResponses, snackbarVariants } from '~/constants'
+import { defaultResponses } from '~/constants'
 import { authRoutes } from '~/router/constants/authRoutes'
 import {
   columns,
@@ -21,15 +22,11 @@ import {
 } from '~/containers/my-resources/lessons-container/LessonsContainer.constants'
 import { type Lesson, ResourcesTabsEnum } from '~/types'
 import { adjustColumns, getScreenBasedLimit } from '~/utils/helper-functions'
-import { useAppDispatch } from '~/hooks/use-redux'
-import { openAlert } from '~/redux/features/snackbarSlice'
-import { getErrorKey } from '~/utils/get-error-key'
 import { useModalContext } from '~/context/modal-context'
 import ChangeResourceConfirmModal from '~/containers/change-resource-confirm-modal/ChangeResourceConfirmModal'
 import { getFullUrl } from '~/utils/get-full-url'
 
 const LessonsContainer = () => {
-  const dispatch = useAppDispatch()
   const navigate = useNavigate()
   const { openModal } = useModalContext()
   const { page, handleChangePage } = usePagination()
@@ -37,6 +34,7 @@ const LessonsContainer = () => {
   const searchTitle = useRef<string>('')
   const breakpoints = useBreakpoints()
   const [selectedItems, setSelectedItems] = useState<string[]>([])
+  const { handleErrorAlert } = useSnackbarAlert()
 
   const { sort } = sortOptions
   const itemsPerPage = getScreenBasedLimit(breakpoints, itemsLoadLimit)
@@ -90,7 +88,7 @@ const LessonsContainer = () => {
       component: (
         <ChangeResourceConfirmModal
           onConfirm={() => {
-            return navigate(
+            navigate(
               getFullUrl({
                 pathname: authRoutes.myResources.editLesson.route,
                 parameters: { id }
@@ -110,14 +108,9 @@ const LessonsContainer = () => {
 
   useEffect(() => {
     if (error) {
-      dispatch(
-        openAlert({
-          severity: snackbarVariants.error,
-          message: getErrorKey(error)
-        })
-      )
+      handleErrorAlert(error)
     }
-  }, [error, dispatch])
+  }, [handleErrorAlert, error])
 
   const props = {
     columns: columnsToShow,
--- a/src/containers/my-resources/lessons-container/LessonsContainer.tsx
+++ b/src/containers/my-resources/lessons-container/LessonsContainer.tsx
@@ -19,22 +19,14 @@ import {
   itemsLoadLimit,
   removeColumnRules
 } from '~/containers/my-resources/lessons-container/LessonsContainer.constants'
-import {
-  ItemsWithCount,
-  Lesson,
-  ErrorResponse,
-  ResourcesTabsEnum
-} from '~/types'
-import {
-  adjustColumns,
-  createUrlPath,
-  getScreenBasedLimit
-} from '~/utils/helper-functions'
+import { type Lesson, ResourcesTabsEnum } from '~/types'
+import { adjustColumns, getScreenBasedLimit } from '~/utils/helper-functions'
 import { useAppDispatch } from '~/hooks/use-redux'
 import { openAlert } from '~/redux/features/snackbarSlice'
 import { getErrorKey } from '~/utils/get-error-key'
 import { useModalContext } from '~/context/modal-context'
 import ChangeResourceConfirmModal from '~/containers/change-resource-confirm-modal/ChangeResourceConfirmModal'
+import { getFullUrl } from '~/utils/get-full-url'
 
 const LessonsContainer = () => {
   const dispatch = useAppDispatch()
@@ -54,17 +46,14 @@ const LessonsContainer = () => {
     removeColumnRules
   )
 
-  const getMyLessons = useCallback(async (): Promise<
-    ItemsWithCount<Lesson>
-  > => {
-    const response = await ResourceService.getUsersLessons({
+  const getMyLessons = useCallback(() => {
+    return ResourceService.getUsersLessonsQuery({
       limit: itemsPerPage,
       skip: (page - 1) * itemsPerPage,
       sort,
       title: searchTitle.current,
       categories: selectedItems
     })
-    return response.data
   }, [page, itemsPerPage, sort, searchTitle, selectedItems])
 
   const deleteLesson = useCallback(
@@ -73,11 +62,10 @@ const LessonsContainer = () => {
   )
 
   const {
-    data: response,
+    data: lessons,
     isLoading,
-    isError,
     error,
-    refetch: fetchData
+    refetch: fetchLessons
   } = useQuery({
     queryKey: [
       'lessons',
@@ -89,18 +77,26 @@ const LessonsContainer = () => {
     ],
     queryFn: getMyLessons,
     options: {
-      initialData: defaultResponses.itemsWithCount
+      staleTime: Infinity
     }
   })
 
   const onEdit = (id: string) => {
-    const resource = response?.items.find((item) => item._id === id)
+    if (!lessons) {
+      return
+    }
+    const resource = lessons.items.find((item) => item._id === id)
     openModal({
       component: (
         <ChangeResourceConfirmModal
-          onConfirm={() =>
-            navigate(createUrlPath(authRoutes.myResources.editLesson.path, id))
-          }
+          onConfirm={() => {
+            return navigate(
+              getFullUrl({
+                pathname: authRoutes.myResources.editLesson.route,
+                parameters: { id }
+              })
+            )
+          }}
           resourceId={id}
           title={resource?.title}
         />
@@ -108,24 +104,27 @@ const LessonsContainer = () => {
     })
   }
 
-  const handleFetchData = useCallback(async () => {
-    await fetchData()
-  }, [fetchData])
+  const handleFetchLessons = useCallback(async () => {
+    await fetchLessons()
+  }, [fetchLessons])
 
   useEffect(() => {
-    if (isError && error) {
+    if (error) {
       dispatch(
         openAlert({
           severity: snackbarVariants.error,
-          message: getErrorKey(error as ErrorResponse)
+          message: getErrorKey(error)
         })
       )
     }
-  }, [isError, error, dispatch])
+  }, [error, dispatch])
 
   const props = {
     columns: columnsToShow,
-    data: { response, getData: handleFetchData },
+    data: {
+      response: lessons ?? defaultResponses.itemsWithCount,
+      getData: handleFetchLessons
+    },
     services: { deleteService: deleteLesson },
     itemsPerPage,
     actions: { onEdit },
@@ -137,16 +136,16 @@ const LessonsContainer = () => {
   return (
     <Box>
       <AddResourceWithInput
-        btnText={'myResourcesPage.lessons.addBtn'}
-        fetchData={handleFetchData}
+        btnText='myResourcesPage.lessons.addBtn'
+        fetchData={handleFetchLessons}
         link={authRoutes.myResources.newLesson.path}
-        placeholder={'myResourcesPage.lessons.searchInput'}
+        placeholder='myResourcesPage.lessons.searchInput'
         searchRef={searchTitle}
         selectedItems={selectedItems}
         setItems={setSelectedItems}
         sortOptions={sortOptions}
       />
-      {isLoading ? (
+      {isLoading || !lessons ? (
         <Loader pageLoad size={50} />
       ) : (
         <MyResourcesTable<Lesson> {...props} />
--- a/src/containers/my-resources/lessons-container/LessonsContainer.tsx
+++ b/src/containers/my-resources/lessons-container/LessonsContainer.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useRef, useState } from 'react'
+import { useCallback, useRef, useState, useEffect } from 'react'
 import { useNavigate } from 'react-router-dom'
 import Box from '@mui/material/Box'
 
@@ -8,7 +8,7 @@ import MyResourcesTable from '~/containers/my-resources/my-resources-table/MyRes
 import Loader from '~/components/loader/Loader'
 import useSort from '~/hooks/table/use-sort'
 import useBreakpoints from '~/hooks/use-breakpoints'
-import useAxios from '~/hooks/use-axios'
+import useQuery from '~/hooks/use-query'
 import usePagination from '~/hooks/table/use-pagination'
 
 import { defaultResponses, snackbarVariants } from '~/constants'
@@ -21,7 +21,6 @@ import {
 } from '~/containers/my-resources/lessons-container/LessonsContainer.constants'
 import {
   ItemsWithCount,
-  GetResourcesParams,
   Lesson,
   ErrorResponse,
   ResourcesTabsEnum
@@ -55,46 +54,47 @@ const LessonsContainer = () => {
     removeColumnRules
   )
 
-  const onResponseError = useCallback(
-    (error?: ErrorResponse) => {
-      dispatch(
-        openAlert({
-          severity: snackbarVariants.error,
-          message: getErrorKey(error)
-        })
-      )
-    },
-    [dispatch]
-  )
-
-  const getMyLessons = useCallback(
-    () =>
-      ResourceService.getUsersLessons({
-        limit: itemsPerPage,
-        skip: (page - 1) * itemsPerPage,
-        sort,
-        title: searchTitle.current,
-        categories: selectedItems
-      }),
-    [page, itemsPerPage, sort, searchTitle, selectedItems]
-  )
+  const getMyLessons = useCallback(async (): Promise<
+    ItemsWithCount<Lesson>
+  > => {
+    const response = await ResourceService.getUsersLessons({
+      limit: itemsPerPage,
+      skip: (page - 1) * itemsPerPage,
+      sort,
+      title: searchTitle.current,
+      categories: selectedItems
+    })
+    return response.data
+  }, [page, itemsPerPage, sort, searchTitle, selectedItems])
 
   const deleteLesson = useCallback(
     (id?: string) => ResourceService.deleteLesson(id ?? ''),
     []
   )
 
-  const { response, loading, fetchData } = useAxios<
-    ItemsWithCount<Lesson>,
-    GetResourcesParams
-  >({
-    service: getMyLessons,
-    defaultResponse: defaultResponses.itemsWithCount,
-    onResponseError
+  const {
+    data: response,
+    isLoading,
+    isError,
+    error,
+    refetch: fetchData
+  } = useQuery({
+    queryKey: [
+      'lessons',
+      page,
+      itemsPerPage,
+      sort,
+      searchTitle.current,
+      selectedItems
+    ],
+    queryFn: getMyLessons,
+    options: {
+      initialData: defaultResponses.itemsWithCount
+    }
   })
 
   const onEdit = (id: string) => {
-    const resource = response.items.find((item) => item._id === id)
+    const resource = response?.items.find((item) => item._id === id)
     openModal({
       component: (
         <ChangeResourceConfirmModal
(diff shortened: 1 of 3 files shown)


src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -21,7 +21,6 @@ import {
 } from '~/containers/my-resources/lessons-container/LessonsContainer.constants'
 import {
   ItemsWithCount,
-  GetResourcesParams,
   Lesson,
   ErrorResponse,
   ResourcesTabsEnum
```
luiqor: Please, add type

```suggestion
  type ItemsWithCount,
  type Lesson,
  type ErrorResponse,
  ResourcesTabsEnum
```

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,46 +54,47 @@ const LessonsContainer = () => {
...
+    ],
+    queryFn: getMyLessons,
+    options: {
+      initialData: defaultResponses.itemsWithCount
+    }
   })
 
   const onEdit = (id: string) => {
-    const resource = response.items.find((item) => item._id === id)
+    const resource = response?.items.find((item) => item._id === id)
```
luiqor: Why did you decide to add `?` ? Don't we always have `response` as the result from server or just initialData

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,46 +54,47 @@ const LessonsContainer = () => {
...
-      ResourceService.getUsersLessons({
-        limit: itemsPerPage,
-        skip: (page - 1) * itemsPerPage,
-        sort,
-        title: searchTitle.current,
-        categories: selectedItems
-      }),
-    [page, itemsPerPage, sort, searchTitle, selectedItems]
-  )
+  const getMyLessons = useCallback(async (): Promise<
```
ShadowOfTheSpace: Why did you typed this callback?

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -108,9 +108,24 @@ const LessonsContainer = () => {
     })
   }
 
+  const handleFetchData = useCallback(async () => {
+    await fetchData()
+  }, [fetchData])
+
+  useEffect(() => {
+    if (isError && error) {
```
ShadowOfTheSpace: `isError` is just `Boolean(error)`, so you can omit it
```suggestion
    if (error) {
```

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -108,9 +108,24 @@ const LessonsContainer = () => {
...
+      dispatch(
+        openAlert({
+          severity: snackbarVariants.error,
+          message: getErrorKey(error as ErrorResponse)
```
ShadowOfTheSpace:
```suggestion
          message: getErrorKey(error)
```

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,46 +54,47 @@ const LessonsContainer = () => {
...
-  >({
-    service: getMyLessons,
-    defaultResponse: defaultResponses.itemsWithCount,
-    onResponseError
+  const {
+    data: response,
+    isLoading,
+    isError,
+    error,
+    refetch: fetchData
```
ShadowOfTheSpace: Please, add meaningful name for that
```suggestion
    refetch: fetchLessons
```

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,46 +54,47 @@ const LessonsContainer = () => {
...
+      page,
+      itemsPerPage,
+      sort,
+      searchTitle.current,
+      selectedItems
+    ],
+    queryFn: getMyLessons,
+    options: {
+      initialData: defaultResponses.itemsWithCount
+    }
```
ShadowOfTheSpace: Try to get rid of `initialData` and use `staleTime` for caching. Without those features `useQuery` is just async wrapper with no additional features at all

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,46 +54,47 @@ const LessonsContainer = () => {
...
+  > => {
+    const response = await ResourceService.getUsersLessons({
+      limit: itemsPerPage,
+      skip: (page - 1) * itemsPerPage,
+      sort,
+      title: searchTitle.current,
+      categories: selectedItems
+    })
+    return response.data
+  }, [page, itemsPerPage, sort, searchTitle, selectedItems])
```
ShadowOfTheSpace: Why you didn't update the service?

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,46 +54,47 @@ const LessonsContainer = () => {
...
 
   const deleteLesson = useCallback(
     (id?: string) => ResourceService.deleteLesson(id ?? ''),
```
ShadowOfTheSpace, line 60: why this `id` is optional?

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -123,15 +138,15 @@ const LessonsContainer = () => {
     <Box>
       <AddResourceWithInput
         btnText={'myResourcesPage.lessons.addBtn'}
```
ShadowOfTheSpace:
```suggestion
        btnText='myResourcesPage.lessons.addBtn'
```

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -123,15 +138,15 @@ const LessonsContainer = () => {
...
-        fetchData={fetchData}
+        fetchData={handleFetchData}
         link={authRoutes.myResources.newLesson.path}
         placeholder={'myResourcesPage.lessons.searchInput'}
```
ShadowOfTheSpace:
```suggestion
        placeholder='myResourcesPage.lessons.searchInput'
```

ShadowOfTheSpace (changes requested):

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: You're right, thanks. Changed.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Done.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Done.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Done.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: That's a good idea. Done.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Maybe in some cases, deleteLesson might be called in a context where id is not necessarily provided, for example, if the function is intended to delete the currently selected item, but there isn't always one.
This is just my opinion, as it was before I started editing, so I left it unchanged. Maybe you know if there are cases like the one I described above in our project or should I delete this possibility of optional id?

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Thanks.Done.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Thanks for the reminder. Replaced getUsersLessons with getUsersLessonsQuery (which was created lately in resource-service.ts)

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Typing this callback ensures that TypeScript knows the return type of 'getMyLessons'. Maybe you now another solution.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: The possibility of optional response is used here to prevent potential runtime errors in case response is undefined or null
In my opinion while we generally expect response to contain data from the server or the initial data, there could be cases where the request fails, the data is not yet available, or the component is in a loading state.
But I agree that it was not the best idea.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
luiqor: @ShadowOfTheSpace , deleteLessons logic wil be replaced as a part of other issue ticket  #3181 and weird approach with `id ?? ''` across all resources containers will be fixed in it too

src/containers/my-resources/lessons-container/LessonsContainer.tsx
luiqor: I used "!" in my PR (attachments) because it's more accurate in this case for my mind. We only use onEdit in the table with resources, so when this function is called, the resources are always loaded (never undefined).
But also if you not sure about safety of approach with "!", I would suggest to replace `response?` (which in case of undefined leads to modals with empty titles) [with "if (!response) return" approach, like Mike did in his PR](https://github.com/ita-social-projects/SpaceToStudy-Client/commit/0e8591d8161f2c54f15c99ac56658f40199f1795#diff-a0d111de9fbab09bab190e81699b8d2852634938e6aedadadb9014254ade94a2L98-R86)

SofiiaYevush: That's a great solution. So I've added `if (!response) { return }` and removed `?`

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Done.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
ShadowOfTheSpace: Thanks for info, I got it

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,62 +46,85 @@ const LessonsContainer = () => {
...
+      'lessons',
+      page,
+      itemsPerPage,
+      sort,
+      searchTitle.current,
+      selectedItems
+    ],
+    queryFn: getMyLessons,
+    options: {
+      staleTime: 0
```
ShadowOfTheSpace: Why 0..?

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,62 +46,85 @@ const LessonsContainer = () => {
...
+    }
+    const resource = lessons.items?.find((item) => item._id === id)
     openModal({
       component: (
         <ChangeResourceConfirmModal
-          onConfirm={() =>
-            navigate(createUrlPath(authRoutes.myResources.editLesson.path, id))
-          }
+          onConfirm={() => {
+            return navigate(
```
ShadowOfTheSpace: You don't need return value of `navigate` function

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,62 +46,85 @@ const LessonsContainer = () => {
...
           resourceId={id}
           title={resource?.title}
         />
       )
     })
   }
 
+  const handleFetchLessons = useCallback(async () => {
+    await fetchLessons()
+  }, [fetchLessons])
```
ShadowOfTheSpace: Try to pass `fetchLessons` directly

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,62 +46,85 @@ const LessonsContainer = () => {
...
+
+  useEffect(() => {
+    if (error) {
+      dispatch(
+        openAlert({
+          severity: snackbarVariants.error,
+          message: getErrorKey(error)
+        })
+      )
```
ShadowOfTheSpace: Check `useSnackbarAlert` hook in our codebase

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
```diff
@@ -1,29 +1,75 @@
...
+  const actual = await vi.importActual('react-router-dom')
+  return {
+    ...actual,
+    useNavigate: () => mockNavigate
+  }
+})
+
+vi.mock('~/utils/get-full-url', () => ({
+  getFullUrl: vi.fn(({ pathname, parameters }) => 
+    `${pathname.replace(':id', parameters.id)}`
```
ShadowOfTheSpace:
```suggestion
    pathname.replace(':id', parameters.id)
```

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
```diff
@@ -35,53 +81,193 @@ const lessonMock = {
...
+    await waitFor(() => {
+      expect(mockOpenModal).toHaveBeenCalled();
+    })
+    
+    const expectedUrl = getFullUrl({
+      pathname: authRoutes.myResources.editLesson.route,
+      parameters: { id: '0' }
+    })
+
+    expect(expectedUrl).toBe(`my-resources/edit-lesson/0`)
```
ShadowOfTheSpace: How this test related to your component?

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
```diff
@@ -35,53 +81,193 @@ const lessonMock = {
...
+    expect(mockDispatch).toHaveBeenCalledWith(
+      openAlert({
+        severity: snackbarVariants.error,
+        message: 'mockedErrorMessage'
+      })
+    )
+  })
+
+  it('should dispatch openAlert when there is an error', async () => {
+    const errorMock = new Error('Test error')
```
ShadowOfTheSpace: This test can't work because you should throw `ResponseError` with all fields instead of plain `Error`.

ShadowOfTheSpace: And same in previous test-case

ShadowOfTheSpace (changes requested):

src/containers/my-resources/lessons-container/LessonsContainer.tsx
ShadowOfTheSpace: Your callback typing isn't a bad idea at all, but in our codebase we don't type return values of all functions (only special ones). So your typed callback looks very strange in codebase.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Because the data has to be updated every time the hook is rerendered. If I use Infinity, when adding a new lesson, the user does not see the added lesson immediately, but only after clicking refresh. Therefore, in this case I need to use 0.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: You're right. Removed.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Great suggestion! Replaced.

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
SofiiaYevush: After clicking on `Edit` button, navigation is performed using `getFullUrl`. So in this test I check whether the url is created as expected. (Maybe the name of the test confused you, there was a typo.)

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
SofiiaYevush: Done.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: It's better to use `handleFetchLessons` because AddResourceWithInput in fetchData and MyResourcesTable expect a different type, which causes errors.
The error occurs because fetchLessons returns a `Promise<QueryObserverResult<ItemsWithCount<Lesson>, ResponseError>>`, while fetchData in AddResourceWithInput expects `() => Promise<void>.`

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
SofiiaYevush: I changed the logic. Thanks!

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
ShadowOfTheSpace: The `getFullUrl` function has already been tested inside its own test file. You only need to test the functionality inside your component. So, I expect you to test just only what is written in your component. Testing something else does not increase your test coverage.
For example, you are using `navigate` inside your component. So you can hardcode the link (or use getFullUrl if you want) and check if `navigate` is called with the particular link that you need.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
ShadowOfTheSpace: `staleTime: 0` is by default in `useQuery`. This means that you have no caching at all. 

For example, if you visit a different page on our app and go back to the lessons page, all your lessons will be downloaded again. We don't need this behavior in our app, we are trying to make fewer redundant requests.

If you need to force `useQuery` to fetch new data after mutation (for example, you can create a new lesson), you need to use `useMutation` and `queryKey` or `queryKeys` to mark some keys as invalid. And after that `useQuery` will make an auto-refetch and you will have fresh data immediately.

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
```diff
@@ -35,53 +95,187 @@ const lessonMock = {
...
   })
 
   it('should render "New lesson" button', () => {
-    const addBtn = screen.getByText('myResourcesPage.lessons.addBtn')
+    useQuery.mockReturnValue({
+      data: lessonResponseMock.items,
+      isLoading: false,
+      error: null,
+      refetch: vi.fn(),
+    })
```
ShadowOfTheSpace: Please, don't mock `useQuery`. Instead use `mockAxiosClient` and mock some return value of your request.

You don't need to do that in all your test cases. Just mock `axiosClient` ones. If you will face any problem, please message me, I'll help you.

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
```diff
@@ -35,53 +95,187 @@ const lessonMock = {
...
+    const expectedUrl = getFullUrl({
+      pathname: authRoutes.myResources.editLesson.route,
+      parameters: { id: '0' }
+    })
 
-    const modal = await screen.findByTestId('confirmModal')
+    expect(expectedUrl).toBe(`my-resources/edit-lesson/0`)
 
-    expect(modal).toBeInTheDocument()
+    mockNavigate(expectedUrl)
```
ShadowOfTheSpace: Why do you calling `navigate` inside your test directly? It should be called in your component, not inside your tests.
There is no testing at all, if you just call something here.

ShadowOfTheSpace (changes requested):

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Investigated and changed. Done. Thanks for the explanation!

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
SofiiaYevush: Done. Thanks for the explanation!

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
SofiiaYevush: Done.

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
SofiiaYevush: Changed.

luiqor (approved):

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,62 +46,79 @@ const LessonsContainer = () => {
...
+      staleTime: Infinity
+    }
   })
 
   const onEdit = (id: string) => {
-    const resource = response.items.find((item) => item._id === id)
+    if (!lessons) {
+      return
+    }
+    const resource = lessons.items?.find((item) => item._id === id)
```
ShadowOfTheSpace: Can `items` be undefined?

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
```diff
@@ -35,52 +65,132 @@ const lessonMock = {
...
     mockAxiosClient
       .onGet(URLs.resources.lessons.get)
       .reply(200, lessonResponseMock)
+      
+      vi.mocked(useNavigate).mockReturnValue(mockNavigate)
+      useQuery.mockReturnValue({
+        data: lessonResponseMock.items,
+        isLoading: false,
+        error: null
+      })
```
ShadowOfTheSpace: Please, format your files

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
ShadowOfTheSpace, line 1: Have you tried to use `mockAxiosClient` instead of mocking `useQuery`?

src/containers/my-resources/lessons-container/LessonsContainer.tsx
```diff
@@ -55,62 +46,79 @@ const LessonsContainer = () => {
...
+    options: {
+      staleTime: Infinity
+    }
   })
 
   const onEdit = (id: string) => {
-    const resource = response.items.find((item) => item._id === id)
+    if (!lessons) {
+      return
+    }
```
ShadowOfTheSpace, line 86: Always add an empty line after `if` statements
```suggestion
    }
    
```

ShadowOfTheSpace (changes requested):

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
SofiiaYevush: Yes, line 82

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Yes, it can.

src/containers/my-resources/lessons-container/LessonsContainer.tsx
SofiiaYevush: Done.

tests/unit/containers/my-resources/LessonsContainer.spec.jsx
ShadowOfTheSpace: Yea, I got it, thanks

src/containers/my-resources/lessons-container/LessonsContainer.tsx
ShadowOfTheSpace: But when?
The backend always returns `count` and `items`.

Probably, you have the wrong mocks there. Check it, please

SofiiaYevush: I checked and can confirm that it can be undefined or null. And besides it's good to have a check to detect any problems/errors in the future.

I can delete `?` here => `const resource = lessons.items?.find((item) => item._id === id)`
and add check there:
  `if (!lessons || !lessons.items) {
    return
  }`
  
  But I think they're the same, so I haven't changed anything...

ShadowOfTheSpace: You have the wrong mocks inside your tests.
Inside your `useQuery` mock you always return `lessonResponseMock.items` instead of `lessonResponseMock` (there is no `items` inside `items` so you have an exception without that `?`. 

About difference between `?` and `if (!lessons || !lessons.items) { return }`
Actually, they are not the same. If you have a return statement, your function will not execute lines after return. If you skip the condition and have `?`, your `resource` will be `undefined` and the confirmation modal will be open without a title (it's not correct logic).

So, please, fix that. That is the reason why you think that `items` can be `undefined`.

ShadowOfTheSpace (approved):

Mike-Popovych (approved):

PR #3219, opened 2025-02-22
Changed the url on the backend so I had to change iet on the frontend

markgol777:
Due to the backend route not following rest api url recommendations I had to change it.

So here's the pr for it on the frontend

### DIFF
--- a/src/constants/request.ts
+++ b/src/constants/request.ts
@@ -120,7 +120,7 @@ export const URLs = {
     add: '/finished-quizzes',
     get: '/finished-quizzes',
     getById: '/finished-quizzes/:id',
-    getByQuizId: '/finished-quizzes/by-quiz-id/:cooperationId/:quizId'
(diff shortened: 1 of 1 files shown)


Made1ra (approved):

ArtemHolikov (approved):

src/constants/request.ts
```diff
@@ -120,7 +120,7 @@ export const URLs = {
     add: '/finished-quizzes',
     get: '/finished-quizzes',
     getById: '/finished-quizzes/:id',
-    getByQuizId: '/finished-quizzes/by-quiz-id/:cooperationId/:quizId'
+    getByQuizId: '/finished-quizzes/:cooperationId/:quizId'
```
ShadowOfTheSpace, line 123: Let's talk about that with the team in our daily meeting. It's not how it should be.

PR #3220, opened 2025-02-23
Created review attempt page and replaced fixed a bug in activeQuiz

### LINKED ISSUE #3170: Create attempt review page
Image

### LINKED ISSUE #3206: Fix an issue about finishing the quiz
Once we'll have the review page instead of having the finish attempt (viewing your mistakes) on the same route we'll just redirect to the freshly created attempt

markgol777:
image

once you click on one of the review attempt pages it takes u to your attempt

image

image

Also when you finish a quiz it takes you to
```http://localhost:3000/my-cooperations/:cooperationid/quizzes/:quizId/attempts/:attemptId```
instead of staying on the same page and changing the state

### CODE REVIEW

File: src/pages/quiz-attempts/QuizAttempts.tsx [hunk 1]
```diff
@@ -59,6 +59,8 @@ const QuizAttemptsPage: React.FC = () => {
     }
   })
 
+  console.log(finishedQuizzes)
```
Made1ra:
```suggestion
```

src/pages/quiz-attempts/QuizAttempts.tsx
```diff
@@ -76,6 +78,15 @@ const QuizAttemptsPage: React.FC = () => {
     setIsOpen(false)
   }
 
+  const handleReviewAttempt = (attemptId: string) => {
```
Made1ra, line 79: Please add 'useCallback' wrapper over this function as it is used as a component prop.

src/pages/quiz/QuizVariants.tsx
```diff
@@ -122,14 +120,15 @@ const ActiveQuiz: React.FC<ActiveQuizProps> = ({ finishQuiz }) => {
     setIsOpen(false)
   }, [])
 
-  const handleFinish = useCallback(async () => {
+  const handleFinish = async () => {
```
Made1ra: Why did you remove `useCallback`?

src/services/base-service.ts
```diff
@@ -31,6 +31,8 @@ export const baseService = {
       return response.data
     } catch (error) {
       if (isAxiosError(error) && error.response) {
+        console.log(error)
```
Made1ra:
```suggestion
```

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
```diff
@@ -56,10 +86,16 @@ describe('QuizPage for student', () => {
         new RegExp(URLs.finishedQuizzes.getById.replace(':id', mockQuizId))
       )
       .reply(200, mockQuiz)
+
+    mockAxiosClient
+      .onGet(
+        '/finished-quizzes//'
```
Made1ra:
```suggestion
        '/finished-quizzes/'
```
I guess there is an extra slash.

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
```diff
@@ -68,4 +104,14 @@ describe('QuizPage for student', () => {
...
   })
+
+  it('should render Quiz review after review button is clicked', async () => {
+    const reviewButton = await screen.findByText('quiz.reviewAttempt')
+
+    act(() => {
+      fireEvent.click(reviewButton)
+    })
+
+    renderWithProviders(<QuizReview />)
```
Made1ra: You should add some `expect` assertions here.

src/pages/quiz-review/QuizReview.tsx
```diff
@@ -0,0 +1,16 @@
+import { useParams } from 'react-router-dom'
+import PageWrapper from '~/components/page-wrapper/PageWrapper'
+import { styles } from '~/pages/quiz-review/QuizPreview.styles'
```
Made1ra: So is it 'review' or 'preview'?

tests/unit/pages/quiz/Quiz.spec.jsx
```diff
@@ -94,6 +95,8 @@ describe('QuizPage for student', () => {
       fireEvent.click(confirmButton)
     })
 
+    renderWithProviders(<QuizReview />)
```
Made1ra: Why do you need it here?

src/pages/quiz/QuizVariants.tsx
Made1ra: Also, you don't need async for this function now.

Made1ra (changes requested):
Please add labels to your pull request.

Made1ra:
> Please add labels to your pull request.

Also, I think the new redirect logic is a new feature, not a bug because we didn't have this route before.

src/pages/quiz-attempts/QuizAttempts.tsx
```diff
@@ -59,6 +59,8 @@ const QuizAttemptsPage: React.FC = () => {
     }
   })
 
+  console.log(finishedQuizzes)
+
   if (isLoading || !quiz) {
```
ArtemHolikov: Are u sure about these 2 conditions? Don't think we need the second, because while we don't have our quiz, isLoading is true and it returns <Loader> component. For !quiz it makes sense to add another condition I guess. For instance, return null If we didn't receive a quiz data.

src/pages/quiz-review/QuizReview.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { FinishedQuiz } from '~/pages/quiz/QuizVariants'
+
+const QuizReview: React.FC = () => {
+  const { attemptId = '' } = useParams()
+
+  return (
+    <PageWrapper sx={styles.quizzesWrapper}>
+      <FinishedQuiz finishedQuizId={attemptId}></FinishedQuiz>
```
ArtemHolikov:
```suggestion
      <FinishedQuiz finishedQuizId={attemptId} />
```

tests/unit/pages/quiz/Quiz.spec.jsx
```diff
@@ -94,6 +95,8 @@ describe('QuizPage for student', () => {
       fireEvent.click(confirmButton)
```
ArtemHolikov:
```suggestion
      fireEvent.click(confirmButton)
```

Why did you use **act** for this button and others? As I know, we don't use it. It's enough to do 

```
fireEvent.click(confirmButton)
``` 
without act

ArtemHolikov (changes requested):

src/pages/quiz-attempts/QuizAttempts.tsx
markgol777: Yes I'm sure about the 2 conditions. I thought you would point out to remove the console log instead of questioning the already working functionality

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
markgol777: No there isn't im not sure why but when I do a console log in base service it does requests to finished-quizzes// and im not sure how to fix it :/

tests/unit/pages/quiz/Quiz.spec.jsx
markgol777: So the tests work

src/pages/quiz-review/QuizReview.tsx
markgol777: Fixed

src/pages/quiz/QuizVariants.tsx
markgol777: - It calls mutateAsync() and navigate(), which don't require referential stability.
- It is already wrapped inside an inline function (onFinish={() => void handleFinish()}), so useCallback would not optimize FinishQuizModal in this case.

src/pages/quiz-attempts/QuizAttempts.tsx
markgol777: Using useCallback Would Be Unnecessary Overhead

If you wrapped handleReviewAttempt in useCallback, you'd still be creating a new arrow function inside .map(), so the optimization would be pointless.

+

there's no expensive computations, navigate is quite fast so I don't see the point of using the hook here

Also I'm calling this function in an inline syntax so even If it caused re-rendering useCallback wouldn't prevent that

markgol777:
> Member

Kinda, I'd label it more as a bug because it was weird that we had a state to check if the quiz is finished. Anyhow it doesn't really matter

src/pages/quiz/QuizVariants.tsx
```diff
@@ -122,14 +120,15 @@ const ActiveQuiz: React.FC<ActiveQuizProps> = ({ finishQuiz }) => {
...
-      navigate(-1)
-    }
-  }, [finishQuiz, mutateAsync, navigate, scoredResponses])
+    navigate(
+      getFullUrl({
+        pathname: authRoutes.cooperationQuizReview.route,
+        parameters: { id: cooperationId, quizId, attemptId: finishedQuiz._id }
+      })
+    )
+  }
```
Made1ra: Please take into account `scoredResponses` setting. I'm not sure that we should show responses in every case. Probably it should be something like this:
```suggestion
  if (scoredResponses) {
      navigate(
        getFullUrl({
          pathname: authRoutes.cooperationQuizReview.route,
          parameters: { id: cooperationId, quizId, attemptId: finishedQuiz._id }
        })
      )
    } else {
      navigate(-1)
    }
  }
```

src/pages/quiz-attempts/QuizAttempts.tsx
ArtemHolikov: Vadym has pointed about console.log, so no point to point out it again. And I just asked about that because thought it's an redudant condition. It can works, but It might work without) It was just a question)

ArtemHolikov (approved):

src/pages/quiz/QuizVariants.tsx
```diff
@@ -122,14 +120,17 @@ const ActiveQuiz: React.FC<ActiveQuizProps> = ({ finishQuiz }) => {
...
-    if (!scoredResponses) {
-      navigate(-1)
+    if (scoredResponses) {
+      navigate(
+        getFullUrl({
+          pathname: authRoutes.cooperationQuizReview.route,
+          parameters: { id: cooperationId, quizId, attemptId: finishedQuiz._id }
+        })
+      )
     }
```
Made1ra: What will happen if the `scoredResponses` is false?

src/pages/quiz-review/QuizReview.tsx
```diff
@@ -0,0 +1,16 @@
...
+import PageWrapper from '~/components/page-wrapper/PageWrapper'
+import { styles } from '~/pages/quiz-review/QuizReview.styles'
+import { FinishedQuiz } from '~/pages/quiz/QuizVariants'
+
+const QuizReview: React.FC = () => {
+  const { attemptId = '' } = useParams()
+
+  return (
+    <PageWrapper sx={styles.quizzesWrapper}>
+      <FinishedQuiz finishedQuizId={attemptId} />
```
ShadowOfTheSpace, line 11: What a point to have a separate component for just wrapping a `FinishedQuiz` and passing an `attemptId`?

Also, I noticed that you are using `useParams` inside `FinishedQuiz` too. So there is no reason why you keep this component separate.

Probably, you should just move all the content of that extra component and move it right here.

src/pages/quiz/QuizVariants.tsx
```diff
@@ -122,14 +120,17 @@ const ActiveQuiz: React.FC<ActiveQuizProps> = ({ finishQuiz }) => {
...
   }, [])
 
-  const handleFinish = useCallback(async () => {
+  const handleFinish = async () => {
     const finishedQuiz = await mutateAsync()
-    setIsOpen(false)
-    finishQuiz(finishedQuiz?._id)
-    if (!scoredResponses) {
-      navigate(-1)
+    if (scoredResponses) {
```
ShadowOfTheSpace, line 171: Always add empty lines around `if` statements. It will improve code readability a lot.
```suggestion
   
    if (scoredResponses) {
```

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
ShadowOfTheSpace: Inside your component, probably, there are `initialValues` that have an `id` as `''`. So instead of a real `id` from mock, it consumes just an empty string as id.
Check that, please, because it's not how it should work.

Maybe you should mock something, I don't know.

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
```diff
@@ -68,4 +104,15 @@ describe('QuizPage for student', () => {
     const quizTitle = await screen.findByText('JS Quiz')
     expect(quizTitle).toBeInTheDocument()
   })
+
+  it('should render Quiz review after review button is clicked', async () => {
+    const reviewButton = await screen.findByText('quiz.reviewAttempt')
+
+    act(() => {
+      fireEvent.click(reviewButton)
+    })
```
ShadowOfTheSpace: Why did you add here `act`, but in other tests you removed it?

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
ShadowOfTheSpace, line 1: Please, don't forget to format your files.

ShadowOfTheSpace (changes requested):

src/pages/quiz-review/QuizReview.tsx
markgol777: I'm not sure about it because we wanted to make finishedQuiz a reusable component also so the tutor can do some grading. Perhaps in the future ill be passing isTutor into it and then conditionally calling it. So I'd rather keep it this way for now

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
markgol777: done

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
markgol777: removed

src/pages/quiz/QuizVariants.tsx
markgol777: fixed

Made1ra (approved):

tests/unit/pages/quiz-attempt/QuizAttempts.spec.jsx
```diff
@@ -56,10 +85,12 @@ describe('QuizPage for student', () => {
         new RegExp(URLs.finishedQuizzes.getById.replace(':id', mockQuizId))
       )
       .reply(200, mockQuiz)
+
+    mockAxiosClient.onGet('/finished-quizzes//').reply(200, mockFinishedQuizzes)
```
ShadowOfTheSpace: It will make this mock more understandable.

```suggestion
    mockAxiosClient.onGet(URLs.finishedQuizzes.getById.replace(':id','')).reply(200, mockFinishedQuizzes)
```
But the better way to handle this is to mock `useParams` with `mockQuizId` as you need.

ShadowOfTheSpace (changes requested):

ShadowOfTheSpace (approved):

PR #3221, opened 2025-02-24
Replaced useAxios to useQuery in the OfferDetails component

### LINKED ISSUE #3215: OfferDetails

ArtemHolikov:
- [x] Replaced useAxios with useQuery on the UserProfile page
- [x] Fixed tests

https://github.com/user-attachments/assets/11b2e7fe-b83f-44e8-94ca-c87a51e3ef63

### DIFF
--- a/src/pages/offer-details/OfferDetails.tsx
+++ b/src/pages/offer-details/OfferDetails.tsx
@@ -27,7 +27,6 @@ import { errorRoutes } from '~/router/constants/errorRoutes'
 import topBlockIcon from '~/assets/img/offer-details/top-block-icon.png'
 import { styles } from '~/pages/offer-details/OfferDetails.styles'
 import {
-  CreateOrUpdateOfferData,
   OutletContext,
   StatusEnum,
   ErrorResponse,
@@ -95,8 +94,7 @@ const OfferDetails = () => {
 
   const { mutate: updateOfferDetails, isPending: updateLoading } = useMutation({
     queryKeys: [['offers'], ['offer', id]],
-    mutationFn: (updateData: Partial<CreateOrUpdateOfferData>) =>
-      OfferService.updateOfferWithBaseService(id, updateData),
+    mutationFn: OfferService.updateOfferWithBaseService,
     onError: responseError
   })
 
--- a/src/pages/offer-details/OfferDetails.tsx
+++ b/src/pages/offer-details/OfferDetails.tsx
@@ -101,7 +101,7 @@ const OfferDetails = () => {
   )
 
   const { mutate: updateOfferDetails, isPending: updateLoading } = useMutation({
-    queryKey: ['offer', id],
+    queryKeys: [['offers'], ['offer', id]],
     mutationFn: updateOffer,
     onError: responseError
   })
--- a/src/pages/offer-details/OfferDetails.tsx
+++ b/src/pages/offer-details/OfferDetails.tsx
@@ -93,16 +93,10 @@ const OfferDetails = () => {
     }
   }, [isError, responseError])
 
-  const updateOffer = useCallback(
-    (updateData: Partial<CreateOrUpdateOfferData>) => {
-      return OfferService.updateOfferWithBaseService(id, updateData)
-    },
-    [id]
-  )
-
   const { mutate: updateOfferDetails, isPending: updateLoading } = useMutation({
     queryKeys: [['offers'], ['offer', id]],
-    mutationFn: updateOffer,
+    mutationFn: (updateData: Partial<CreateOrUpdateOfferData>) =>
+      OfferService.updateOfferWithBaseService(id, updateData),
     onError: responseError
   })
 
@@ -149,7 +143,7 @@ const OfferDetails = () => {
         : StatusEnum.Draft
 
     if (offerData) {
-      updateOfferDetails({ status })
+      updateOfferDetails({ status, id })
     }
   }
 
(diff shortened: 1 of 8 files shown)


src/pages/offer-details/OfferDetails.tsx
```diff
@@ -202,16 +205,16 @@ const OfferDetails = () => {
...
+  }, [dispatch, isOfferLoading])
 
   useEffect(() => {
     void dispatch(
       fetchUserById({ userId, role: userRole as UserRole, isEdit: false })
     )
   }, [dispatch, userId, userRole])
 
-  if (offerLoading) {
+  if (isOfferLoading) {
```
ShadowOfTheSpace: You should add `offerData` to the condition.
`isOfferLoading` is a condition when `queryFn` is "active"
`offerData` is for type safety. Without adding it to the condition, you will have TS errors, when you will use offer data directly. Also, there is no reason to show `null` instead of loader, when the offer is not defined. Why do we need layout shifts because of `null`?

tests/unit/pages/offer-details/OfferDetails.spec.jsx
```diff
@@ -278,32 +269,4 @@ describe('Offer details with student role', () => {
...
-    }
-    renderWithProviders(<OfferDetails />, {
-      preloadedState: newMockState
-    })
-
-    const loader = screen.getByTestId('loader')
-
-    expect(loader).toBeInTheDocument()
-  })
-})
```
ShadowOfTheSpace, line 309: Why did you delete this test? You can mock `useQuery` inside this test and try to find a loader on your page.

src/pages/offer-details/OfferDetails.tsx
```diff
@@ -69,35 +69,41 @@ const OfferDetails = () => {
...
+      responseError()
+    }
+  }, [isError, responseError])
+
   const updateOffer = useCallback(
-    (updateData?: Partial<CreateOrUpdateOfferData>) =>
-      OfferService.updateOffer(id, updateData),
+    (updateData: Partial<CreateOrUpdateOfferData>) => {
+      return OfferService.updateOfferWithBaseService(id, updateData)
+    },
```
ShadowOfTheSpace: You can remove this wrapper and use `mutate` function directly.

src/pages/offer-details/OfferDetails.tsx
```diff
@@ -69,35 +69,41 @@ const OfferDetails = () => {
...
-    null,
-    Partial<CreateOrUpdateOfferData>
-  >({
-    service: updateOffer,
-    fetchOnMount: false,
-    defaultResponse: null,
-    onResponseError: responseError
+  const { mutate: updateOfferDetails, isPending: updateLoading } = useMutation({
+    queryKeys: [['offers'], ['offer', id]],
+    mutationFn: updateOffer,
```
ShadowOfTheSpace: And instead of this callback, as a `mutationFn` you can pass just a service `OfferService.updateOfferWithBaseService`

src/pages/offer-details/OfferDetails.tsx
```diff
@@ -136,36 +142,33 @@ const OfferDetails = () => {
...
+  const handleToggleOfferStatus = () => {
     const status =
       offerData?.status === StatusEnum.Draft
         ? StatusEnum.Active
         : StatusEnum.Draft
 
     if (offerData) {
-      await fetchDataUpdateOffer({ status })
-      void fetchDataOffer()
+      updateOfferDetails({ status })
```
ShadowOfTheSpace: Don't forget to pass the `id` to the parameters of `updateOfferDetails` mutation function in all its usages.

ShadowOfTheSpace (changes requested):

tests/unit/pages/offer-details/OfferDetails.spec.jsx
ArtemHolikov: Discussed it with another teammate that it doesn't make sense to test loader. But okay, I'm on my way to fix it!

src/pages/offer-details/OfferDetails.tsx
ShadowOfTheSpace: It's my mistake, sorry, you cannot pass service directly if it contains more than 1 parameter. Let's leave that callback here.

src/pages/offer-details/OfferDetails.tsx
ShadowOfTheSpace: Previous comment

src/pages/offer-details/OfferDetails.tsx
ShadowOfTheSpace: Previous comment

ShadowOfTheSpace (approved):

Mike-Popovych (approved):

src/pages/offer-details/OfferDetails.tsx
```diff
@@ -69,35 +68,34 @@ const OfferDetails = () => {
...
-    Partial<CreateOrUpdateOfferData>
-  >({
-    service: updateOffer,
-    fetchOnMount: false,
-    defaultResponse: null,
-    onResponseError: responseError
+  const { mutate: updateOfferDetails, isPending: updateLoading } = useMutation({
+    queryKeys: [['offers'], ['offer', id]],
+    mutationFn: OfferService.updateOfferWithBaseService,
+    onError: responseError
```
SofiiaYevush, line 98: Shouldn't we add onSuccess here?

SofiiaYevush, line 98: Also in `onError` it's better to import `useSnackbarAlert from '~/hooks/use-snackbar-alert'` and use `handleErrorAlert` from `useSnackbarAlert` instead of `responseError`. 
Please check it.

SofiiaYevush (changes requested):

src/pages/offer-details/OfferDetails.tsx
ArtemHolikov: We don't need it, because it has already success alert after offer details update. Here I provided a video to demonstrate how it works:

https://github.com/user-attachments/assets/2552fa2b-a041-4914-86a7-07b30a85a9df

ArtemHolikov: Yes, imported and used, but in another place, thank you. We can't use it instead of ```responseError``` because they have 2 different meaning. ```responseError``` is responsible for redirecting user to errorPage in case of some errors.

PR #3271, opened 2025-03-15
Replaced useAxios with useMutation in useSteps hook

### LINKED ISSUE #3237: (SP: 2) Replace `useAxios` with `useMutation` in the `useSteps` hook
- Replace useAxios with useMutation.
- Update error handling.
- Update service.
- Update tests and add a few if neccessary.

SofiiaYevush:
- Replaced `useAxios` with `useMutation` for improved API request handling.
- Extracted error handling into `useSnackbarAlert` for better separation of concerns.
- Refactored `updateUser` service to simplify logic.
- Replaced `axiosClient.patch` with `baseService.request` for a more flexible and consistent API service structure.
- Improved readability and maintainability of the `useSteps` hook.
- Updated tests in `EditProfileSlice.spec.js` file.

Using the hook:
image
(Stepper after the first user registration)

Successfully created PATCH request:
image

**Issue: #3237**

### DIFF
--- a/src/hooks/use-steps.tsx
+++ b/src/hooks/use-steps.tsx
@@ -5,7 +5,12 @@ import { useModalContext } from '~/context/modal-context'
 import { useStepContext } from '~/context/step-context'
 import { userService } from '~/services/user-service'
 import { snackbarVariants } from '~/constants'
-import { type StepData, type UpdateUserParams } from '~/types'
+import {
+  type EditProfilePhoto,
+  type UpdatedPhoto,
+  type StepData,
+  type UpdateUserParams
+} from '~/types'
 import { openAlert } from '~/redux/features/snackbarSlice'
 import useSnackbarAlert from '~/hooks/use-snackbar-alert'
 
@@ -74,8 +79,23 @@ const useSteps = ({ steps }: UseSteps) => {
     const { firstName, lastName, country, city, professionalSummary } =
       stepData.generalInfo.data
 
+    let formattedPhoto: EditProfilePhoto | undefined = undefined
+
+    if (stepData.photo && stepData.photo.length > 0) {
+      const firstPhoto = stepData.photo[0]
+
+      if (firstPhoto instanceof File) {
+        formattedPhoto = {
+          src: URL.createObjectURL(firstPhoto),
+          name: firstPhoto.name
+        }
+      } else {
+        formattedPhoto = firstPhoto as UpdatedPhoto
+      }
+    }
+
     const data: UpdateUserParams = {
-      photo: stepData.photo[0] ? stepData.photo[0] : '',
+      photo: formattedPhoto,
       firstName,
       lastName,
       address: {
@@ -101,7 +121,13 @@ const useSteps = ({ steps }: UseSteps) => {
     setActiveStep
   }
 
-  return { activeStep, stepErrors, isLastStep, stepOperation, loading }
+  return {
+    activeStep,
+    stepErrors,
+    isLastStep,
+    stepOperation,
+    loading
+  }
 }
 
 export default useSteps
--- a/src/hooks/use-steps.tsx
+++ b/src/hooks/use-steps.tsx
@@ -5,12 +5,7 @@ import { useModalContext } from '~/context/modal-context'
 import { useStepContext } from '~/context/step-context'
 import { userService } from '~/services/user-service'
 import { snackbarVariants } from '~/constants'
-import {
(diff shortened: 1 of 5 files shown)


src/services/user-service.tsx
```diff
@@ -45,20 +44,21 @@ export const userService = {
...
     })
   },
-
-  updateUser: (
-    userId: string,
-    params: UpdateUserParams
-  ): Promise<AxiosResponse<null>> => {
-    return axiosClient.patch(createUrlPath(URLs.users.update, userId), params)
+  updateUser: (userId: string, params: UpdateUserParams) => {
+    return baseService.request<null>({
```
ArtemHolikov:
```suggestion
    return baseService.request<void>({
```

src/services/user-service.tsx
```diff
@@ -45,20 +44,21 @@ export const userService = {
...
+      method: 'PATCH',
+      url: getFullUrl({
+        pathname: `${URLs.users.update}/${userId}`
```
ArtemHolikov, line 51:
```suggestion
        pathname: URLs.users.update,
        parameters: { id: userId }
```
and change ```update``` URL in ```request.ts``` file to: '/users/:id'

Can we make so?

src/redux/features/editProfileSlice.ts
```diff
@@ -154,10 +155,12 @@ export const updateUser = createAsyncThunk(
   ) => {
     try {
       const response = await userService.updateUser(userId, params)
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
ArtemHolikov, line 158: Maybe we can just check for type somehow to avoid this?

ArtemHolikov (changes requested):

src/redux/features/editProfileSlice.ts
SofiiaYevush: Yes. There is another issue for this (#3284 ). So it's just a temporary solution.

src/services/user-service.tsx
SofiiaYevush: I tried this, but it will not work correct, because `update` endpoint don't have an `id`.

src/services/user-service.tsx
SofiiaYevush: I will investigate it.
The reason why I used there `null` is because we used it in AxiosResponse: `Promise<AxiosResponse<null>>`

SofiiaYevush: Replaced. It was great idea) I replaced `null` with `void` to keep logic safe.

ArtemHolikov (approved):

markgol777 (approved):

src/hooks/use-steps.tsx
```diff
@@ -21,21 +19,15 @@ const useSteps = ({ steps }: UseSteps) => {
...
   const dispatch = useAppDispatch()
   const { userId } = useAppSelector((state) => state.appMain)
+  const { handleErrorAlert } = useSnackbarAlert()
 
-  const updateUser = useCallback(
-    (params?: UpdateUserParams) => userService.updateUser(userId, params!),
+  const handleUpdateUser = useCallback(
+    (params?: UpdateUserParams) => {
+      return userService.updateUser(userId, params!)
+    },
```
Mike-Popovych, line 27:
```suggestion
  const handleUpdateUser = useCallback(
    (params: UpdateUserParams) => {
      return userService.updateUser(userId, params)
    },
```

We were using both params?: UpdateUserParams and params! in the same function, which is contradictory. The ? indicates that the argument is optional, while the ! forcefully asserts that it's defined.
Since we always expect params to be passed, it's cleaner and safer to make it a required parameter and remove the unnecessary non-null assertion.

src/services/user-service.tsx
Mike-Popovych: We can add /:id to our user request – this will solve the issue since the user is updated only by ID. That way, we don’t need a separate /users/ request. I’d avoid concatenation here if possible, but up to you.

Mike-Popovych (approved):

src/hooks/use-steps.tsx
SofiiaYevush: Thanks for advice) I will fix it in the next issue.

src/services/user-service.tsx
SofiiaYevush: Thanks) I will try to change it in the next issue.

PR #3276, opened 2025-03-19
Replaced useAxios with useQuery and useMutation in CreateCourseContainer

### LINKED ISSUE #3261: Replace useAxios with useQuery in CreateCourse

Mike-Popovych:
- [X] Replaced useAxios with useQuery and useMutation in CreateCourseContainer
 - [X] Removed  eslint-disable-next-line react-hooks/exhaustive-deps
 - [X] Changed tests
image

https://github.com/user-attachments/assets/ab54c847-852d-44bf-bafe-33cc99734595


src/constants/request.ts
```diff
@@ -40,9 +40,10 @@ export const URLs = {
   },
   courses: {
     get: '/courses',
+    getById: '/courses/:id',
     delete: '/courses',
     create: '/courses',
-    patch: '/courses'
+    patch: '/courses/:id'
```
ArtemHolikov, line 46: I would rename it to "update", but it's up to you.
```suggestion
    update: '/courses/:id'
```

ArtemHolikov (approved):

src/constants/request.ts
Mike-Popovych: Personally, I think patch is better here, because update could also mean a PUT method. This way, it avoids confusion in the services.

VKormylo (approved):

SofiiaYevush (approved):

PR #3317, opened 2025-04-04
Fixed error in editProfileSlice

### LINKED ISSUE #3284: (SP: 1) Fix error in editProfileSlice
Fix error in editProfileSlice

Path:
Image

Code:
Image

Error:
Image

SofiiaYevush:
- Removed `response.data` return. Now `updateUser` service returns void (no data expected from PATCH requests), so we simply await completion.
- Explicit `return true` on success. Replaced unsafe type-ignored returns with a clear success flag for better intent and type safety.
- Updated endpoint and service.
- Updated test.

Examples of successful updating of the data:
image

image

image

image

**Issue: #3284**

### DIFF
--- a/src/redux/features/editProfileSlice.ts
+++ b/src/redux/features/editProfileSlice.ts
@@ -155,7 +155,6 @@ export const updateUser = createAsyncThunk(
   ) => {
     try {
       await userService.updateUser(userId, params)
-      return
     } catch (e) {
       if (e instanceof ResponseError) {
         return rejectWithValue(e.code)
--- a/src/redux/features/editProfileSlice.ts
+++ b/src/redux/features/editProfileSlice.ts
@@ -155,7 +155,7 @@ export const updateUser = createAsyncThunk(
   ) => {
     try {
       await userService.updateUser(userId, params)
-      return true
+      return
     } catch (e) {
       if (e instanceof ResponseError) {
         return rejectWithValue(e.code)
--- a/src/redux/features/editProfileSlice.ts
+++ b/src/redux/features/editProfileSlice.ts
@@ -157,10 +157,7 @@ export const updateUser = createAsyncThunk(
       await userService.updateUser(userId, params)
       return true
     } catch (e) {
-      const error = e as AxiosError<ErrorResponse>
-      if (error.response?.data?.code) {
-        return rejectWithValue(error.response.data.code)
-      } else if (e instanceof ResponseError) {
+      if (e instanceof ResponseError) {
         return rejectWithValue(e.code)
       }
     }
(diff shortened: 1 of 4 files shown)


ArtemHolikov (commented):
image

Cover it by tests, please

src/redux/features/editProfileSlice.ts
```diff
@@ -154,9 +154,8 @@ export const updateUser = createAsyncThunk(
     { rejectWithValue }
   ) => {
     try {
-      const response = await userService.updateUser(userId, params)
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-      return response.data
+      await userService.updateUser(userId, params)
+      return true
```
ArtemHolikov:
```suggestion
      await userService.updateUser(userId, params)
```

I don't think we need ```return true``` here. For what needs? We don't have to return true because you call the update function and it shouldn't return nothing. Or I am mistaken?

ArtemHolikov (changes requested):

ArtemHolikov (changes requested):
image

And cover your code by tests, please

src/redux/features/editProfileSlice.ts
SofiiaYevush: My intention was to return `true` as a successful ending of `updateUser`, and for `fulfilled` state. 
But if you think it is unnecessary, I will test updating without `true`. It's good advice

ArtemHolikov: It will update user in successful ending. In case you get an error, ```catch``` block catchs it and returns with ```rejectWithValue```

SofiiaYevush:
> image
> 
> And cover your code by tests, please

Covered)

src/redux/features/editProfileSlice.ts
SofiiaYevush: You're right, that's true. Changed

src/redux/features/editProfileSlice.ts
```diff
@@ -154,9 +154,8 @@ export const updateUser = createAsyncThunk(
     { rejectWithValue }
   ) => {
     try {
-      const response = await userService.updateUser(userId, params)
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-      return response.data
+      await userService.updateUser(userId, params)
+      return
```
ArtemHolikov:
```suggestion

```

I mean, u don't need ```return``` in general. Yes, it says that it completely finished, but it's just from stylistically side, no more.

ArtemHolikov (changes requested):

nebby2105 (approved):

src/redux/features/editProfileSlice.ts
SofiiaYevush: You're absolutely right) Changed

ArtemHolikov (approved):

markgol777 (approved):

Mav-Ivan (approved):

PR #3343, opened 2025-04-12
Added ua translation for categories and subjects

### LINKED ISSUE #3324: (SP: 2) Add ukrainian translations for categories and subjects in pages: Categories, Subjects, Find Offers
Add ukrainian translations for categories and subjects. 
**It relates to Categories page, Subjects page and Find Offers page.** 
It should be working similar to translation of categories in Subjects page.

**Here is how it currently works in Subjects page (also update incorrect translations):**

Image

Image

Mav-Ivan:
https://github.com/user-attachments/assets/9752219c-076b-43e5-8d9b-71ca0523ab49


Mike-Popovych (approved):

VKormylo (changes requested):
In Subjects page and AddProfessionalCategoryModal translateData function is still used for translating categories and subjects. I think it would be better if you replace that with your newly created useTranslate hook.

It looks something like this:
image

Mav-Ivan:
> In Subjects page and AddProfessionalCategoryModal translateData function is still used for translating categories and subjects. I think it would be better if you replace that with your newly created useTranslate hook.
> 
> It looks something like this: image

Thanks for the suggestion, I’ve replaced it with the useTranslate hook.

VKormylo (approved):

nebby2105 (approved):

PR #3443, opened 2025-05-13
Fix profile update to reflect changes without page reload 

### LINKED ISSUE #3442: (SP: ) Fix profile update to reflect changes without page reload
When we update the name and last name, and then go to the profile page 
Image

We see the old data 
Image

Only after manual reload we can see the new one 
Image

uliaescha:
Before: 

https://github.com/user-attachments/assets/ca113a8a-4e4e-42c3-a36d-261e44f101a0

- Rewrote user profile update logic using `useMutation` instead of  `dispatch`
- Fixed navigation after update
- Fixed modal confirmation logic: after clicking the Update button, the confirmation dialog is no longer triggered unnecessarily  
- Fixed tests 

After: 

https://github.com/user-attachments/assets/d42b8482-3b19-497b-ada6-05869e41cd9d

### DIFF
--- a/src/pages/edit-profile/EditProfile.tsx
+++ b/src/pages/edit-profile/EditProfile.tsx
@@ -13,6 +13,8 @@ import Box from '@mui/material/Box'
 import Divider from '@mui/material/Divider'
 
 import useConfirm from '~/hooks/use-confirm'
+import useMutation from '~/hooks/use-mutation'
+import useSnackbarAlert from '~/hooks/use-snackbar-alert'
 import { useAppDispatch, useAppSelector } from '~/hooks/use-redux'
 import Loader from '~/components/loader/Loader'
 import PageWrapper from '~/components/page-wrapper/PageWrapper'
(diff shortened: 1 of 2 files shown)


lizabre (approved):

danikua (approved):

SofiiaYevush (commented):
Is it the right solution after clicking on the `Update` button return to the previous `My profile` page?

uliaescha:
> Is it the right solution after clicking on the `Update` button return to the previous `My profile` page?

In the code, it was like `if (hash) { navigate(`${authRoutes.myProfile.path}#complete`) }` before, but the redirect didn't work at all. So I fixed that issue.

SofiiaYevush:
> > Is it the right solution after clicking on the `Update` button return to the previous `My profile` page?
> 
> In the code, it was like `if (hash) { navigate(`${authRoutes.myProfile.path}#complete`) }` before, but the redirect didn't work at all. So I fixed that issue.

Got it.

SofiiaYevush (approved):

